PR #11, opened 2018-09-11
Add rule for services modifying their internal state

dnna:
With the arrival of different execution models like [PHP-PM](https://github.com/php-pm/php-pm) and [Roadrunner](https://github.com/spiral/roadrunner), the container and related services can no longer be guaranteed to be destroyed on every request. A common source of bugs in these scenarios, are services that modify their state (for example changing internal attributes when a method is called) without resetting it at the end of the request, assuming the service will be destroyed and reset.

Static analysis presents a good opportunity to catch these bugs and help make applications or libraries compatible with these models. For this reason I'm proposing to add a rule that detects changes to a service's internal state, if its outside the constructor or the method injections defined in the config.

### DIFF
--- a/extension.neon
+++ b/extension.neon
@@ -20,6 +20,10 @@ services:
 		tags:
 			- phpstan.rules.rule
 	-
+		class: PHPStan\Rules\Symfony\ModifiedServiceStateRule
+		tags:
+			- phpstan.rules.rule
+	-
 		class: PHPStan\Type\Symfony\ContainerInterfaceMethodTypeSpecifyingExtension
 		tags:
 			- phpstan.typeSpecifier.methodTypeSpecifyingExtension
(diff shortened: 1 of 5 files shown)


lookyman:
This is a VERY interesting PR, however I feel this would deserve it's own extension. I might incorporate some of the changes to facilitate easier extending of Service and/or ServiceMap classes with third party code. Stay tuned. 😊

lookyman:
I've made some refactoring in 0.10.2:

* `Service` now implements the `ServiceDefinition` interface.
* `ServiceMap` now takes an array of `ServiceDefinition`s in the constructor.
* The services are created by a factory implementing the `ServiceMapFactory` interface.
* The service map factory service is named in the extension.

Altogether this should allow you to override the service map factory with your own, returning your own implementation of the service definition interface, and do whatever you want from there. Let me know if you run into any issues.

dnna:
Thanks @lookyman ! I will look into it and try to create an extension when I have some time 😊

PR #14, opened 2018-10-24
Use spaces instead of tabs in config example

robertfausk:
of phpstan.neon cause tabs are not allowed and will result in an error after c&p this example and running phpstan.

Error message example:
```
Note: Using configuration file /var/www/html/phpstan.neon.dist.

In Decoder.php line 357:
                                                               
  Invalid combination of tabs and spaces on line 7, column 9. 
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,10 +26,10 @@ And include extension.neon in your project's PHPStan config:
 
 ```
 includes:
-	- vendor/phpstan/phpstan-symfony/extension.neon
+    - vendor/phpstan/phpstan-symfony/extension.neon
 parameters:
-	symfony:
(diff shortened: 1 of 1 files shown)


ondrejmirtes:
In NEON tabs are allowed but you can't mix them with spaces.

ondrejmirtes:
So it doesn't make sense to merge this, someone might bump into an opposite issue.

robertfausk:
What do you think about leaving a hint next to this example that copy & paste can lead to an error (as happened to me)?

ondrejmirtes:
I think the message is self-evident and is easy to fix so there doesn’t have to be a notice in the README.

PR #19, opened 2018-11-09
Disable PrivateServiceRule when the class implements ServiceSubscriberInterface

iamluc:
I have a lot of violations `Service "X" is private.` when extending the `AbstractController` which implements the `ServiceSubscriberInterface`.

When implementing this interface, the service receive a special container with the services listed by the method `getSubscribedServices`.

I think the PrivateServiceRule should be disabled in this case.

(For details: https://symfony.com/doc/current/service_container/service_subscribers_locators.html)


lookyman:
Merged as  c340de8, thanks!

PR #34, opened 2019-02-13
Add support for Serializer::deserialize

mcfedr:
The return type of deserialize depends on the type you ask it for, so this gets us the correct type.

### DIFF
--- a/src/Type/Symfony/SerializerInterfaceDynamicReturnTypeExtension.php
+++ b/src/Type/Symfony/SerializerInterfaceDynamicReturnTypeExtension.php
@@ -1,0 +1,45 @@
+<?php declare(strict_types = 1);
+
+namespace PHPStan\Type\Symfony;
+
+use PhpParser\Node\Expr\MethodCall;
+use PHPStan\Analyser\Scope;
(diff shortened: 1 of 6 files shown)


lookyman:
Thanks!

src/Type/Symfony/SerializerInterfaceDynamicReturnTypeExtension.php
```diff
@@ -0,0 +1,45 @@
...
+		$argType = $scope->getType($methodCall->args[1]->value);
+		if (!$argType instanceof ConstantStringType) {
+			return new MixedType();
+		}
+
+		$objectName = $argType->getValue();
+
+		if (substr($objectName, -2) === '[]') {
+			// The key type is determined by the data
+			return new ArrayType(new MixedType(false), new ObjectType(substr($objectName, 0, -2)));
```
jvasseur: This part sould be recursive as you can define types like this `Foo[][]` and the denormalizer will handle them.

mcfedr: You are right of course, https://github.com/phpstan/phpstan-symfony/pull/35

PR #35, opened 2019-02-13
Add support for recursive types to Serializer

mcfedr:
As correctly pointed out by @jvasseur, it's possible to have arrays of arrays.

https://github.com/phpstan/phpstan-symfony/pull/34#discussion_r256559742

Of course really we should check what normalizers are in the container, and only if ArrayNormalizer is active support this, and possibly be able to say what objects are deserializable,


lookyman:
We can cross that bridge when we get to it (e.g. when someone runs into that and opens an issue 😊).

mcfedr:
> We can cross that bridge when we get to it (e.g. when someone runs into that and opens an issue 😊).

Yes, when I stated writing this i thought about it, and decided it was extra complication for very little gain...

Issue #27, opened 2018-12-06
False positive in Tests: Service XY is private


Pierstoval:
When using the `KernelTestCase` and asking PhpStan to analyze a test, I have this error:

![image](https://user-images.githubusercontent.com/3369266/49603146-bc9cdd00-f98a-11e8-9e7d-6eff7c8cbe69.png)

However, the `static::$container` that is in the `KernelTestCase` is supposed to be an instance of the `TestContainer` but is typed as `ContainerInterface` because it might happen that the `test.container` service can be absent from the container, therefore making `static::$container` be the full container (as of [KernelTestCase here](https://github.com/symfony/symfony/blob/8f36bbd/src/Symfony/Bundle/FrameworkBundle/Test/KernelTestCase.php#L87) )

Here is the code that triggers the error: [Pierstoval/AgateApps/.../RedeemerTest.php#L61-L79](https://github.com/Pierstoval/AgateApps/blob/aa3cd13bb4f496efd47fd722ab6941b7404a041d/tests/Voucher/Redeem/RedeemerTest.php#L61-L79)

If the `test.container` service exists in the container AND for this specific case, I think there shouldn't be any error.

lookyman:
I plan to try to tackle these false positives in the near future. It's... difficult. Symfony container is not as straightforward as I imagined when I started working with it and writing this extension. 😊

Pierstoval:
By chance there are not _that many_ errors with the container 👍

Wirone:
I was struggling with this too, but @Pierstoval's solution worked for me. Maybe it's not as clean as it should be, but resolves my CI issue :) Subscribing to this, so hopefully it can be replaced with proper solution on some point.

mitelg:
plus 1 :blush: 
would be nice, if this could be fixed somehow

@Pierstoval's solution also worked for me in the meantime :+1:

jeroennoten:
I created #210 to resolve this issue.

ondrejmirtes:
Fixed by: https://github.com/phpstan/phpstan-symfony/pull/210

Merged and released :)

Pierstoval:
By the way, the exact same will happen to `KernelTestCase::getContainer()` method, because using `self::$container` is now deprecated as of Symfony 5.3 😅

ondrejmirtes:
Feel free to send another PR :)

Pierstoval:
[...]

PR #39, opened 2019-02-26
Separate rules from extension as in other phpstan extensions

mcfedr:
Doing this because issue of false positives in test cases, https://github.com/phpstan/phpstan-symfony/issues/27, but anyway, its the style in all the other phpstan plugins, so would be good to do the same here

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,14 @@ parameters:
 		container_xml_path: '%rootDir%/../../../var/cache/dev/srcApp_KernelDevDebugContainer.xml' 
 ```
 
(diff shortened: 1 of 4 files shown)


ondrejmirtes:
I agree, but README also needs updating and this can be released as 0.12 because it's a BC break.

mcfedr:
Added README changes

ondrejmirtes:
This needs rebasing because of new rules and conflict in extension.neon

lookyman:
Thanks!

PR #41, opened 2019-03-05
Fix issue while loading a container

jdeniau:
When the URI of the container contains escaped characters like `%2F`, `simplexml_load_file` throw an error.

~According [to the documentation](http://php.net/manual/en/function.simplexml-load-file.php) :~
> ~Libxml 2 unescapes the URI, so if you want to pass e.g. b&c as the URI parameter a, you have to call simplexml_load_file(rawurlencode('http://example.com/?a=' . urlencode('b&c'))). Since PHP 5.1.0 you don't need to do this because PHP will do it for you.~

~I don't understand exactly what should not be done since PHP 5.1.0, but with PHP 7.2, I reproduced the following error :~
```php
$jd = 'jdjd/test.xml';
$jdSlash = rawurlencode('jdjd/test.xml');
$jnSlash = rawurlencode('jn%2Fjn/test.xml');
$jn = 'jn%2Fjn/test.xml';

var_dump($jd); // string(13) "jdjd/test.xml"
var_dump(simplexml_load_file($jd)); // OK object(SimpleXMLElement) 

var_dump($jdSlash); // string(15) "jdjd%2Ftest.xml"
var_dump(simplexml_load_file($jdSlash)); // OK object(SimpleXMLElement) 

var_dump($jnSlash); // string(20) "jn%252Fjn%2Ftest.xml"
var_dump(simplexml_load_file($jnSlash)); // OK object(SimpleXMLElement)

var_dump($jn); // string(16) "jn%2Fjn/test.xml"
var_dump(simplexml_load_file($jn)); // PHP Warning:  simplexml_load_file(): I/O warning : failed to load external entity "jn%2Fjn/test.xml"
```

Another possible fix is to use:
```php
$xml = @simplexml_load_string(file_get_contents($this->containerXml));
```

### DIFF
--- a/src/Symfony/XmlServiceMapFactory.php
+++ b/src/Symfony/XmlServiceMapFactory.php
@@ -2,7 +2,7 @@
 
 namespace PHPStan\Symfony;
 
-use function simplexml_load_file;
+use function simplexml_load_string;
 use function sprintf;
 use function strpos;
 use function substr;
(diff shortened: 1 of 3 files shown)


lookyman:
I like the `simplexml_load_string` solution better. Also, could you add a test for that? Thanks!

jdeniau:
@lookyman done

lookyman:
Thank you!

PR #42, opened 2019-03-12
remove case sensitivity from service ids

### LINKED ISSUE #9: services should not be case sensitive
it is common pattern to name services as class path, and then select them with `SomeClass::class`.

symfony converts those dependencies to lower case before saving / resolving.

This causing errors like `Service "SomeNamespace\SomeClass" is not registered in the container.`

because in `appDevDebugProjectContainer.xml` it is saved as: `somenamespace\someclass`

stefanotorresi:
fixes #9

### DIFF
--- a/src/Symfony/ServiceMap.php
+++ b/src/Symfony/ServiceMap.php
@@ -18,7 +18,7 @@ final class ServiceMap
 	 */
 	public function __construct(array $services)
 	{
-		$this->services = $services;
+		$this->services = array_change_key_case($services);
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


lookyman:
What version of Symfony are you using?

lookyman:
closing, see https://github.com/phpstan/phpstan-symfony/issues/9#issuecomment-420616940

stefanotorresi:
that's too bad: I think the issue is not in the upstream, because service definitions are NOT case sensitive. It just happens that newer version of Symfony generate the xml manifest with the original case instead of applying `strtolower`.

lookyman:
Ok, I'll reconsider it...

jvasseur:
> because service definitions are NOT case sensitive.
That's false, services id are case sensitive since Symfony 4.

stefanotorresi:
> That's false, services id are case sensitive since Symfony 4.

I see. That's not the case for Symfony 3, still.

stefanotorresi:
I guess I'm gonna have to give up on this as it's not forward compatible with SF4 and granting compatibility with both is out of the scope of my original intent.
Thanks for your input, @jvasseur!

PR #44, opened 2019-03-28
Add EnvelopeReturnTypeExtension for symfony/messenger

### LINKED ISSUE #43: False positive with Symfony Messenger
`Symfony\Component\Messenger\Envelope::all()` returns a [different type](https://github.com/symfony/messenger/blob/9db2bb6af3e1bf12de6bf4d5dcdf250c8642adcc/Envelope.php#L74-L84) based on whether or not an argument was passed to it. Currently PHPStan doesn't know that which results in errors like this:

```
Parameter #1 ... of method ... expects array<array<Symfony\Component\Messenger\Stamp\StampInterface>>, array<array<Symfony\Component\Messenger\Stamp\StampInterface>|Symfony\Component\Messenger\Stamp\StampInterface> given.
```

I'd send PR but I'm not sure how cases like these should be solved in PHPStan. If you can give me an advice how to do it, I'll send the PR.

enumag:
Closes #43

### DIFF
--- a/src/Type/Symfony/EnvelopeReturnTypeExtension.php
+++ b/src/Type/Symfony/EnvelopeReturnTypeExtension.php
@@ -1,0 +1,47 @@
+<?php declare(strict_types = 1);
+
+namespace PHPStan\Type\Symfony;
+
+use PhpParser\Node\Expr\MethodCall;
+use PHPStan\Analyser\Scope;
+use PHPStan\Reflection\MethodReflection;
+use PHPStan\Type\ArrayType;
+use PHPStan\Type\Constant\ConstantStringType;
+use PHPStan\Type\DynamicMethodReturnTypeExtension;
+use PHPStan\Type\MixedType;
+use PHPStan\Type\ObjectType;
+use PHPStan\Type\Type;
+use Symfony\Component\Messenger\Stamp\StampInterface;
+
+final class EnvelopeReturnTypeExtension implements DynamicMethodReturnTypeExtension
+{
+
+	public function getClass(): string
+	{
+		return 'Symfony\Component\Messenger\Envelope';
+	}
+
+	public function isMethodSupported(MethodReflection $methodReflection): bool
+	{
+		return $methodReflection->getName() === 'all';
+	}
+
+	public function getTypeFromMethodCall(
+		MethodReflection $methodReflection,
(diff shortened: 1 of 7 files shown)


tests/Type/Symfony/envelope_all.php
```diff
@@ -0,0 +1,9 @@
+<?php declare(strict_types = 1);
+
+$envelope = new \Symfony\Component\Messenger\Envelope(new stdClass());
+
+$test1 = $envelope->all(\Symfony\Component\Messenger\Stamp\ReceivedStamp::class);
+$test2 = $envelope->all(random_bytes(1));
```
enumag: any better way to pass a "non-constant-string" in this test?

lookyman: I guess this is fine.

src/Type/Symfony/EnvelopeReturnTypeExtension.php
```diff
@@ -0,0 +1,47 @@
...
+	}
+
+	public function getTypeFromMethodCall(
+		MethodReflection $methodReflection,
+		MethodCall $methodCall,
+		Scope $scope
+	): Type
+	{
+		if (count($methodCall->args) === 0) {
+			return new ArrayType(new MixedType(), new ArrayType(new MixedType(), new ObjectType(StampInterface::class)));
```
lookyman: `StampInterface` needs to be a string here, so that composer require checker doesn't complain. I know, not ideal.

src/Type/Symfony/EnvelopeReturnTypeExtension.php
```diff
@@ -0,0 +1,47 @@
...
+		}
+
+		$argType = $scope->getType($methodCall->args[0]->value);
+		if (!$argType instanceof ConstantStringType) {
+			return new ArrayType(new MixedType(), new ObjectType(StampInterface::class));
```
lookyman: ...and here...

enumag:
@lookyman Fixed. In my opinion composer require checker should ignore `::class` usages since they don't fail if the class/interface does not exist.

lookyman:
@enumag Yes, I think so as well.

Some things are still failing, but I'll fix those. Thanks!

enumag:
@lookyman Thank you!

PR #45, opened 2019-04-09
Console commands

### LINKED ISSUE #20: Help getArgument/getOption in Symfony Console InputInterface
Symfony had a few recent PRs to the Symfony console that impacted my PHPStan analysis due to PHPDoc changes:

https://github.com/symfony/symfony/pull/28448
https://github.com/symfony/symfony/pull/28374
https://github.com/symfony/symfony/pull/28647

Bottom line:
 * `InputInterface::getArgument` now returns `string|string[]|null`
 * `InputInterface::getOption` returns `string|string[]|bool|null`. 

In reality, they can return `string[]` only if the relative argument/option is set as `InputArgument::IS_ARRAY`/`InputOption::VALUE_IS_ARRAY`.

Is it possible to add an aid in this package for this case? The minimum repro case is this one:

```php
class PhpstanCommand extends \Symfony\Component\Console\Command\Command
{
    protected function configure()
    {
        $this->addArgument('argument', InputArgument::REQUIRED);
        $this->addOption('option', null, InputOption::VALUE_REQUIRED);
    }

    protected function execute(InputInterface $input, OutputInterface $output)
    {
        // Cannot cast array<string>|string|null to string. 
        $argument = (string) $input->getArgument('argument');
        // Cannot cast array<string>|bool|string|null to string. 
        $option = (string) $input->getOption('option');
    }
}
```

### LINKED ISSUE #36: [Command] getOption return types
This is a question, I'd be interested in implementing this, but not sure if its possible with Phpstan, so a point in the right direction would be useful.

In Symfony commands options (and arguments) are added in `configure` function, and this affects the return type of `$input->getOption` later.

```php
class MyCommand {
    
    protected function configure(): void
    {
        $this
            ->addOption('str', null, InputOption::VALUE_REQUIRED, 'This option will be ?string')
            ->addOption('bool', null, InputOption::VALUE_NONE, 'this option will be bool')
            ->addOption('arra', null, InputOption::VALUE_IS_ARRAY, 'this option will be ?(string[])')
        ;
    }

    protected function execute(InputInterface $input, OutputInterface $output): int
    {
        $str = $input->getOption('str');
    }
}
```

So can I change the return type of `getOption`, according to the options that are configured?

lookyman:
closes #20, closes #36

@mcfedr @Jean85 Can you guys test this?

### DIFF
--- a/tests/Rules/Symfony/InvalidOptionDefaultValueRuleTest.php
+++ b/tests/Rules/Symfony/InvalidOptionDefaultValueRuleTest.php
@@ -1,0 +1,35 @@
+<?php declare(strict_types = 1);
+
+namespace PHPStan\Rules\Symfony;
+
+use PHPStan\Rules\Rule;
+use PHPStan\Testing\RuleTestCase;
+
+final class InvalidOptionDefaultValueRuleTest extends RuleTestCase
+{
+
+	protected function getRule(): Rule
+	{
+		return new InvalidOptionDefaultValueRule();
+	}
+
+	public function testGetArgument(): void
+	{
+		$this->analyse(
+			[
+				__DIR__ . '/ExampleCommand.php',
+			],
+			[
+				[
+					'Parameter #5 $default of method Symfony\Component\Console\Command\Command::addOption() expects array<int, string>|null, array<int, int> given.',
+					29,
+				],
+				[
+					'Parameter #5 $default of method Symfony\Component\Console\Command\Command::addOption() expects string|false|null, int given.',
+					30,
+				],
+			]
+		);
+	}
+
+}
--- a/tests/Rules/Symfony/InvalidOptionDefaultValueRuleTest.php
+++ b/tests/Rules/Symfony/InvalidOptionDefaultValueRuleTest.php
@@ -24,10 +24,6 @@ final class InvalidOptionDefaultValueRuleTest extends RuleTestCase
 					'Parameter #5 $default of method Symfony\Component\Console\Command\Command::addOption() expects array<int, string>|null, array<int, int> given.',
 					29,
 				],
-				[
-					'Parameter #5 $default of method Symfony\Component\Console\Command\Command::addOption() expects string|false|null, int given.',
-					30,
-				],
 			]
 		);
 	}
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ This extension provides following features:
 * Provides correct return type for `Envelope::all()` method based on the `$stampFqcn` parameter.
 * Notifies you when you try to get an unregistered service from the container.
 * Notifies you when you try to get a private service from the container.
-* Optionally correct return types for `InputInterface::getArgument()` and `::getOption`
+* Optionally correct return types for `InputInterface::getArgument()`, `::getOption`, `::hasArgument`, and `::hasOption`.
 
 ## Usage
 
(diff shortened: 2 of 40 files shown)


ondrejmirtes:
❤️ ❤️ ❤️

ondrejmirtes:
I'd like the `symfony.container_xml_path` to be nullable - we're using Symfony Console but without the Symfony DIC.

ondrejmirtes:
Basically the extension related to DIC should not do anything if this parameter is null :)

mhujer:
@lookyman Just tested it. Works great 👍 

```
Command "demo:command" does not define argument "foobar".
```

tests/Rules/Symfony/InvalidOptionDefaultValueRuleTest.php
```diff
@@ -0,0 +1,35 @@
...
+			[
+				__DIR__ . '/ExampleCommand.php',
+			],
+			[
+				[
+					'Parameter #5 $default of method Symfony\Component\Console\Command\Command::addOption() expects array<int, string>|null, array<int, int> given.',
+					29,
+				],
+				[
+					'Parameter #5 $default of method Symfony\Component\Console\Command\Command::addOption() expects string|false|null, int given.',
```
mcfedr: This appears to be wrong - the doc is 

`* @param string|string[]|int|bool|null $default     The default value (must be null for InputOption::VALUE_NONE)`

so i'm seeing this reported for my calls to to `addOption` where I'm passing `int`

mcfedr: Although its true that `getOption` has return type of `string|string[]|bool|null` - which will likely also be wrong in this case... maybe the symfony docblock needs to be fixed

mcfedr: the better reference, https://github.com/phpstan/phpstan-symfony/pull/45/files#diff-e486177dda816bfb1e25cf9f96abb146R69

lookyman: The thing is, if you pass the option from the console, you can only ever get string there (or bool in case of VALUE_NONE, or null if you don't pass the option at all). You can pass a different type as a default, but that is imho wrong (as suggested by the docblock), and there is a rule that should forbid that.

lookyman: So there are two options here: 1) go with the docblock and just restrict it's allowed types based on the mode, or 2) union with the default's type, even if it's not allowed by the docblock. I went with the first approach. Do you suggest we should go with the second?

mcfedr: Interestingly this was just recently changed in Symfony - https://github.com/symfony/symfony/pull/29201 - seems there is a bit of inconsistency here in the docblocks and the code, the `ArgvInput`, normally used will only ever give you strings, but the `ArrayInput`, often used for testing, lets you pass `int` (or whatever else).

I guess given the reality is your going to get strings out for real usage, might be best to stick with what you've done. Might be also be nice to check the args to `ArrayInput` are also strings([]) so users dont test with different types.

lookyman: You know what, I actually for some reason haven't noticed that the `addOption` method allows `int` in default. I guess I should reflect that after all.

lookyman: @mcfedr done

lookyman:
> 
> 
> I'd like the `symfony.container_xml_path` to be nullable - we're using Symfony Console but without the Symfony DIC.

> 
> 
> Basically the extension related to DIC should not do anything if this parameter is null :)

@ondrejmirtes done

PR #52, opened 2019-06-21
Alternative way to initialize ConsoleApplicationResolver

ossinkine:
I propose to add an alternative more simple way to initialize `ConsoleApplicationResolver`. The most popular way to create a console application is already described in README - it's just create an application kernel and create a console application with it. And no need to create a file specially for this.

### DIFF
--- a/extension.neon
+++ b/extension.neon
@@ -3,12 +3,14 @@ parameters:
 		container_xml_path: null
 		constant_hassers: true
 		console_application_loader: null
+		console_application_kernel_class: null
 
 parametersSchema:
 	symfony: structure([
 		container_xml_path: schema(string(), nullable())
 		constant_hassers: bool()
 		console_application_loader: schema(string(), nullable())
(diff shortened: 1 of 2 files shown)


ossinkine:
@lookyman WDYT?

lookyman:
TBH I don't like it, but I don't know why. 🤷 Let me get a second opinion. @ondrejmirtes Do we want it?

ondrejmirtes:
I don't think it's necessary. There's already a way to do it which is straightforward. I'm not even sure how this should work - do Symfony application even extend the Kernel class with their own? I worry that people would shoot themselves in the foot with this one and complain it doesn't work.

ossinkine:
I agree that current implementation is more flexable, you can define your application as you mind. But I think a separate file with a console application creation for PHPStan is redundantly. I would to get it simpler, ideally the application from `bin/console` should be used but I don't know how. :)

ondrejmirtes:
I think it’s simple enough without any downside currently.

PR #58, opened 2019-08-02
Request::getSession type specifier.

### LINKED ISSUE #56: Request::getSession and null return value.
Request's method `getSession` can return `null` value, so when I tried call code like this: `$request->getSession()->set(...)` I got error: 

> Cannot call method set() on Symfony\Component\HttpFoundation\Session\SessionInterface|null.

I thought that checking if a request has sessions before call setter, resolve the problem, but unfortunately not. I created a repo which gives You a possibility to reproduce the problem:

https://github.com/plotek/symfony-phpstan

Possible workaround:
```php
if ($request->hasSession()) {
    $session = $request->getSession();
    Assert::isInstanceOf($session, SessionInterface::class);
    $session->set('Test', 'Value');
}
```

plotek:
Closes #56

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -26,8 +26,7 @@
     "php": "^7.1",
     "ext-simplexml": "*",
     "phpstan/phpstan": "^0.11.7",
-    "nikic/php-parser": "^4.0",
-    "symfony/http-foundation": "^3.0 || ^4.0"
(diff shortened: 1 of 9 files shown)


lookyman:
Thanks!

PR #60, opened 2019-08-16
Symfony Configuration

tomasfejfar:
Ref.: https://github.com/phpstan/phpstan/issues/844


tomasfejfar:
@lookyman could you help me a little with how it's supposed to work? I thought that I add expressions and it checks that they are what I expect they are. But unless I do what's in 8941637 it fails with "true is not false" (i.e. the $run equals false) :(

ondrejmirtes:
Superseded by https://github.com/phpstan/phpstan-symfony/pull/126

PR #63, opened 2019-11-01
Parse container for Psr\Container\ContainerInterface as well

### LINKED ISSUE #26: PSR-11 Container and AbstractController support
https://github.com/symfony/framework-bundle/blob/master/Controller/Controller.php is marked as deprecated since 4.2 and there is a new way of defining a controller through https://github.com/symfony/framework-bundle/blob/master/Controller/AbstractController.php. 

It also uses PSR-11 Container interface https://github.com/php-fig/container/blob/master/src/ContainerInterface.php, so currently phpstan not being able to find services.

tomasfejfar:
Fixes #26


lookyman:
Merged as 18887de5a3334b306307ad214d83be0a79857f28, thanks!

PR #68, opened 2019-12-10
Add stubs for (some) Symfony iterable types

VasekPurchart:
These are types where PHPStan on lvl 6 was complaining about missing iterable information (in our codebase) - there will be probably more, but I think especially the form ones are quite wide spread.

Original types for reference:

* [FormBuilderInterface](https://github.com/symfony/symfony/blob/v5.0.1/src/Symfony/Component/Form/FormBuilderInterface.php)
* [FormInterface](https://github.com/symfony/symfony/blob/v5.0.1/src/Symfony/Component/Form/FormInterface.php)
* [Session](https://github.com/symfony/symfony/blob/v5.0.1/src/Symfony/Component/HttpFoundation/Session/Session.php)
* [HeaderBag](https://github.com/symfony/symfony/blob/v5.0.1/src/Symfony/Component/HttpFoundation/HeaderBag.php)

As per discussion with @ondrejmirtes I am not adding any dependencies.

### DIFF
--- a/stubs/HeaderBag.php
+++ b/stubs/HeaderBag.php
@@ -1,0 +1,16 @@
+<?php
+
+namespace Symfony\Component\HttpFoundation;
+
+/**
+ * @implements \IteratorAggregate<string, string|string[]>
+ */
+class HeaderBag implements \IteratorAggregate
+{
+
+	/**
+	 * @phpstan-return \Traversable<string, string|string[]>
+	 */
+	public function getIterator() {}
+
+}
--- a/stubs/Session.php
+++ b/stubs/Session.php
@@ -1,0 +1,16 @@
+<?php
+
+namespace Symfony\Component\HttpFoundation\Session;
+
+/**
+ * @implements \IteratorAggregate<string, mixed>
+ */
+class Session implements \IteratorAggregate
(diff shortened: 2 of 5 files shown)


stubs/HeaderBag.php
```diff
@@ -0,0 +1,16 @@
+<?php
+
+namespace Symfony\Component\HttpFoundation;
+
+/**
+ * @implements \Traversable<string, string|string[]>
```
ondrejmirtes: This is wrong - it has to `@implements \IteratorAggregate`

stubs/Session.php
```diff
@@ -0,0 +1,16 @@
+<?php
+
+namespace Symfony\Component\HttpFoundation\Session;
+
+/**
+ * @implements \Traversable<string, mixed>
```
ondrejmirtes: This is wrong - it has to `@implements \IteratorAggregate`

ondrejmirtes:
Thanks!

VasekPurchart:
Thanks, tested dev-master with deleted stubs in our codebase and works as expected :)

lookyman:
Thank you!

PR #72, opened 2020-01-20
Add a bunch of stubs

dbrekelmans:
Mostly adding `array<mixed>` where symfony documents `array` (phpstan will complain about this on max level when implementing these interfaces). 

Also added an "empty" `ContainerBuilder` stub, since phpstan complained about not knowing the class when seeing it in `ExtensionInterface.stub`. If there's another way to fix this please let me know.

These changes are based off of symfony 4.4 and tested on a medium to large-ish codebase.

### DIFF
--- a/stubs/EventSubscriberInterface.stub
+++ b/stubs/EventSubscriberInterface.stub
@@ -1,0 +1,11 @@
+<?php
+
+namespace Symfony\Component\EventDispatcher;
+
+interface EventSubscriberInterface
+{
+    /**
+     * @return array<string, string|array<string, int>|array<int, string|array<string, int>|array<int, string>>>
+     */
+    public static function getSubscribedEvents();
+}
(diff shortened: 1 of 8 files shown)


stubs/EventSubscriberInterface.stub
```diff
@@ -0,0 +1,11 @@
+<?php
+
+namespace Symfony\Component\EventDispatcher;
+
+interface EventSubscriberInterface
+{
+    /**
+     * @return array<string, string|array<string, int>|array<int, string|array<string, int>|array<int, string>>>
```
dbrekelmans: This is clearly documented at https://github.com/symfony/symfony/blob/master/src/Symfony/Component/EventDispatcher/EventSubscriberInterface.php and tested agains a codebase that uses all the different structures.

lookyman:
I'll have a proper look a bit later, but it looks good. Thanks.

lookyman:
Thank you!

PR #87, opened 2020-06-15
Add stub for KernelTestCase

Khartir:
This adds a stub for `Symfony\Bundle\FrameworkBundle\Test\KernelTestCase` to set the type of `$container` to `Symfony\Bundle\FrameworkBundle\Test\TestContainer` which in turn allows access to private services in tests as mentioned in https://github.com/phpstan/phpstan-symfony/issues/4


Khartir:
Apparently this has been fixed.

PR #90, opened 2020-08-11
Update README.md

antishov:
Hi
There is no need to use `%rootDir%/../../../var/cache`, because `%kernel.cache_dir%` parameter is available


ondrejmirtes:
Hi, you're mixing up config of Symfony and PHPStan. PHPStan has no config parameter called `%kernel.cache_dir%`.

PR #92, opened 2020-08-13
Add template for AclProviderInterface

VincentLanglet:
Hi @ondrejmirtes 

If I understand correctly template and I didn't make any mistake, this PR seems better than https://github.com/symfony/security-acl/pull/60.

By the way I have two others questions:
- WDYT about supporting directory in the `stubFiles` config ? This way the following code would be enough.
```
stubFiles:
    - stubs/
```

- WDYT about using the same directory structure than Symfony inside the stub directory ? There will be more and more file, and it can be easier to find them if we follow the namespace.


stubs/AclProviderInterface.stub
```diff
@@ -0,0 +1,21 @@
+<?php
+
+namespace Symfony\Component\Security\Acl\Model;
+
+use Symfony\Component\Security\Acl\Model\AclInterface;
+
+/**
+ * @template TAcl of AclInterface
+ */
+interface AclProviderInterface extends AclProviderInterface
```
ondrejmirtes, line 10: `AclProviderInterface extends AclProviderInterface`?

ondrejmirtes: And BTW you don't have to solve this with generics, I'm not sure it makes sense here. You can simply override the method in the child MutableAclProviderInterface and say it returns MutableAclInterface.

VincentLanglet: > And BTW you don't have to solve this with generics, I'm not sure it makes sense here. You can simply override the method in the child MutableAclProviderInterface and say it returns MutableAclInterface.

I made a new try.

ondrejmirtes: Yeah, this is nicer, thanks. Please fix the build, you need to follow this tidbit:

> Why is PHPStan complaining about “Class not found” in a stub file?

https://phpstan.org/user-guide/stub-files

ondrejmirtes: Yeah, namespaces could be honored in the directory structure :)

For the directories it'd have to be a new config key `stubDirectories` but I don't think it's really necessary.

VincentLanglet: I made the following changes:
- honored namespaces with directory inside the stubs one
- run the linter on the stubs files.
- Add the missing stubs classes.

Not every methods are defined in a stubs file. In the same way, do I have to write `interface MutableAclProviderInterface extends AclProviderInterface` in a stub file or can I just write 
`interface MutableAclProviderInterface` and avoid defining `AclProviderInterface` ?

ondrejmirtes: The build is still failing. It's better to write `interface MutableAclProviderInterface extends AclProviderInterface`, seems more complete, although that information is currently unused from the stubs.

VincentLanglet:
I need https://github.com/phpstan/phpstan/issues/3111 first

VincentLanglet:
@ondrejmirtes I removed the `findAcls` annotation since it requires the next phpstan major.
So this PR is now ready to merge.

VincentLanglet:
@ondrejmirtes I rebase master ; this PR is ready to merge

ondrejmirtes:
Hi, I'm sorry, but this PR does too many things at once and it's unlikely to be merged.

1) Adds AclProviderInterface stub - should be in a separate PR.
2) Moves stubs into subdirectories - should be a separate PR.
3) Adds CS to stubs - not interested in this, sorry, as other repositories don't have this and there's little value in that.

Issue #91, opened 2020-08-11
http-foundation InputBag::get with default value should not return null


raalderink:
If you use HTTPFoundation's Request, and then get specific input bags (query/request,...), the `get` may return `string|null`. However, if you pass parameter #2 `$default` as non-null value, it will never return `null`, and always a string.

So for example this code can never return null;

```
$request->query->get('name', 'default value');
```

Currently, because the typehint on the function is `string|null`, PHPStan will complain if you pass this to a function with a non-null argument, even though it will never return null;

```
Parameter #2 $name of class Foo constructor expects string, string|null given
```

This is tested on Symfony 5.1.

raalderink:
I'll be working on a fix for this and will create a PR when ready

ondrejmirtes:
This needs a dynamic return type extension https://phpstan.org/developing-extensions/dynamic-return-type-extensions :)

ruudk:
Fixed in https://github.com/phpstan/phpstan-symfony/pull/107

Can be closed.

PR #93, opened 2020-08-14
Fix InputBagDynamicReturnTypeExtension

raalderink:
For the InputBag::get, if you add a default string parameter, it will always return a string. This is new in Symfony 5.0.

Fixes issue #91

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -33,7 +33,7 @@
     "squizlabs/php_codesniffer": "^3.3.2",
     "symfony/console": "^4.0",
     "symfony/framework-bundle": "^4.0",
-    "symfony/http-foundation": "^4.0|^5.0",
+    "symfony/http-foundation": "^4.0 || ^5.0",
     "symfony/messenger": "^4.2",
     "symfony/serializer": "^4.0"
   },
--- a/composer.json
+++ b/composer.json
@@ -33,7 +33,7 @@
     "squizlabs/php_codesniffer": "^3.3.2",
     "symfony/console": "^4.0",
     "symfony/framework-bundle": "^4.0",
-    "symfony/http-foundation": "^4.0",
+    "symfony/http-foundation": "^4.0|^5.0",
     "symfony/messenger": "^4.2",
     "symfony/serializer": "^4.0"
   },
--- a/src/Type/Symfony/InputBagDynamicReturnTypeExtension.php
+++ b/src/Type/Symfony/InputBagDynamicReturnTypeExtension.php
@@ -1,0 +1,38 @@
+<?php declare(strict_types=1);
+
+namespace PHPStan\Type\Symfony;
+
+use PhpParser\Node\Expr\MethodCall;
+use PHPStan\Analyser\Scope;
+use PHPStan\Reflection\MethodReflection;
+use PHPStan\Reflection\ParametersAcceptorSelector;
+use PHPStan\Type\Constant\ConstantStringType;
+use PHPStan\Type\DynamicMethodReturnTypeExtension;
+use PHPStan\Type\StringType;
+use PHPStan\Type\Type;
+
+final class InputBagDynamicReturnTypeExtension implements DynamicMethodReturnTypeExtension
+{
+	public function getClass(): string
+	{
+		return 'Symfony\Component\HttpFoundation\InputBag';
+	}
+
+	public function isMethodSupported(MethodReflection $methodReflection): bool
+	{
+		return $methodReflection->getName() === 'get';
+	}
+
+	public function getTypeFromMethodCall(
+		MethodReflection $methodReflection,
+		MethodCall $methodCall,
+		Scope $scope
+	): Type
+	{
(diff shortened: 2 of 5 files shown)


src/Type/Symfony/InputBagDynamicReturnTypeExtension.php
```diff
@@ -0,0 +1,38 @@
...
+use PhpParser\Node\Expr\MethodCall;
+use PHPStan\Analyser\Scope;
+use PHPStan\Reflection\MethodReflection;
+use PHPStan\Reflection\ParametersAcceptorSelector;
+use PHPStan\Type\Constant\ConstantStringType;
+use PHPStan\Type\DynamicMethodReturnTypeExtension;
+use PHPStan\Type\StringType;
+use PHPStan\Type\Type;
+
+final class InputBagDynamicReturnTypeExtension implements DynamicMethodReturnTypeExtension
```
ondrejmirtes, line 14: The extension needs to be registered in `extension.neon`.

composer.json
```diff
@@ -33,7 +33,7 @@
     "squizlabs/php_codesniffer": "^3.3.2",
     "symfony/console": "^4.0",
     "symfony/framework-bundle": "^4.0",
-    "symfony/http-foundation": "^4.0",
+    "symfony/http-foundation": "^4.0|^5.0",
```
ondrejmirtes, line 36: Why this change? Doesn't seem related to the rest.

src/Type/Symfony/InputBagDynamicReturnTypeExtension.php
```diff
@@ -0,0 +1,38 @@
...
+		return $methodReflection->getName() === 'get';
+	}
+
+	public function getTypeFromMethodCall(
+		MethodReflection $methodReflection,
+		MethodCall $methodCall,
+		Scope $scope
+	): Type
+	{
+		if (isset($methodCall->args[1]) && $scope->getType($methodCall->args[1]->value) instanceof ConstantStringType) {
```
ondrejmirtes, line 32: I'd rather use this logic:

```php
$argType = $scope->getType($methodCall->args[1]->value);
$isNull = (new NullType())->isSuperTypeOf($argType);
$isString = (new StringType())->isSuperTypeOf($argType);
$compare = $isNull->compareTo($isString);
if ($compare === $isString) {
    return new StringType();
}

return ParametersAcceptorSelector::selectSingle($methodReflection->getVariants())->getReturnType();
```

It will work better in some edge cases:

1) If we have `mixed`, the default return type will be used.
2) If we have `string|null`, the default return type will be used.

composer.json
raalderink: The InputBag is new in Symfony 5, didn't exist in 4 yet, so to run the tests we need Symfony 5

ondrejmirtes: There currently isn't a nice way to test it in a 3rd party package, so you can delete the file `tests/Type/Symfony/input_bag_get.php` which isn't very useful currently.

raalderink: Okay, fair, so just delete the tests altogether?

ondrejmirtes: Oh sorry, I haven't noticed there's actually a way here in phpstan-symfony with the ExtensionTestCase :) I just thought you create the file `input_bag_get.php` and let PHPStan analyse it, but right now you're actually checking the return types which is great :) So feel free to add some tests that fail with `mixed`, `string|null` etc. and see how my suggested logic fixes that :)

raalderink: I'm not quite sure what you want me too add there? The `$default` parameter can technically be mixed, however, anything other than a string will throw a deprecation notice. Besides, falling back on default will say `string|null` either way, no matter what you put in `$default`, so I'm not sure if it would matter. I'm fine with adding other types to the test but the result will be the same :P

ondrejmirtes: `mixed` means that it can be a string or null and we can't be sure, so the return type also has to be `string|null`. I just want to test the extension for `mixed` and `string|null` types as the second parameter.

raalderink: PR updated, I think that's what you were looking for?

ondrejmirtes: Approximately :) Thanks.

ondrejmirtes: Please fix the build: https://travis-ci.com/github/phpstan/phpstan-symfony/jobs/372130355

ruudk:
@raalderink Anything I can do to help this PR get merged? I'm having the same issue while upgrading to Symfony 5.

ruudk:
Created a PR on your fork to get rid of the failing test. https://github.com/raalderink/phpstan-symfony/pull/1

raalderink:
@ruudk Thanks, I had lost track of this as I went on holiday

ondrejmirtes:
Hi, the logic in the extension is solid but:

1) I don't get the change in composer.json.
2) The build failed.

ruudk:
@raalderink Are you going to fix this or should I take it over?

ruudk:
@ondrejmirtes @raalderink Let's continue in https://github.com/phpstan/phpstan-symfony/pull/107 which is ✅  and ready to merge 🎉

PR #95, opened 2020-08-21
Adjusts getOption return type if no default value.

ddebin:
I read discussion in #45 about adding `int` type for `getOption`. But reading Symfony source code and doing some tests, `int` type is only possible if an `int` default value has been passed to `addOption`.

### DIFF
--- a/src/Type/Symfony/InputInterfaceGetOptionDynamicReturnTypeExtension.php
+++ b/src/Type/Symfony/InputInterfaceGetOptionDynamicReturnTypeExtension.php
@@ -67,12 +67,12 @@ final class InputInterfaceGetOptionDynamicReturnTypeExtension implements Dynamic
 				if (!$option->acceptValue()) {
 					$optType = new BooleanType();
 				} else {
-					$optType = TypeCombinator::union(new StringType(), new IntegerType(), new NullType());
+					$optType = TypeCombinator::union(new StringType(), new NullType());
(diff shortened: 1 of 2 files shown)


ondrejmirtes:
Hi, this doesn't look like the right solution. With your modification, `getOption()` will never contain `int` but in reality it can (in case the default value is null).

ddebin:
> Hi, this doesn't look like the right solution. With your modification, `getOption()` will never contain `int` but in reality it can (in case the default value is null).

@ondrejmirtes No that's right. You cannot get an `int` from `getOption` unless you pass an `int` as default value.

ondrejmirtes:
Yeah but with your changes, PHPStan will never think it’s an int, right?

ddebin:
I'm not sure if I understand your reply but `$optType = TypeCombinator::union($optType, $scope->getTypeFromValue($option->getDefault()));` adds `int` type if default value is an `int`.

ondrejmirtes:
Oh I haven’t noticed that. Is this covered by a test?

ddebin:
> Oh I haven’t noticed that. Is this covered by a test?

Yes with untouched lines L35 and L36 of `InputInterfaceGetOptionDynamicReturnTypeExtensionTest.php`.

ondrejmirtes:
Perfect, I'll release it then. Thank you!

ondrejmirtes:
Tagged as 0.12.8.

PR #100, opened 2020-10-03
Update README.md

ThomasLandauer:
Fixing path. If the file is in `test`, you just need to go up *one* level. See https://github.com/phpstan/phpstan-doctrine/pull/148 for Doctrine.


ondrejmirtes:
Thank you.

PR #106, opened 2020-10-21
Bump minimum versions to prevent deprecations while running in lowest mode

ruudk:
This removes the deprecations. These are confusing for contributors.


ondrejmirtes:
Thank you!

PR #108, opened 2020-10-22
Add FormTypeExtensionInterface stub

pscheit-lillydoo:
Since it implements nearly the same interface, but isn't typehinted either ($options missing iterable type)

How can I test that?

### DIFF
--- a/stubs/FormTypeExtensionInterface.stub
+++ b/stubs/FormTypeExtensionInterface.stub
@@ -1,0 +1,21 @@
+<?php
+
+namespace Symfony\Component\Form;
+
+interface FormTypeExtensionInterface
+{
+    /**
+     * @param array<mixed> $options
+     */
+    public function buildForm(FormBuilderInterface $builder, array $options): void;
+
+    /**
+     * @param array<mixed> $options
+     */
+    public function buildView(FormView $view, FormInterface $form, array $options): void;
+
+    /**
+     * @param array<mixed> $options
+     */
+    public function finishView(FormView $view, FormInterface $form, array $options): void;
+}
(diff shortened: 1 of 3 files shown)


pscheit-lillydoo:
Thank you for writing phpstan!

I accept the code of conduct in phpstan/phpstan-src

Greetings

ondrejmirtes:
Hi, what kind of bug this solves? Thank you.

pscheit-lillydoo:
```php

Class MyTypeExtension implements FormTypeExtensionInterface
{
    public function buildForm(FormBuilderInterface $builder, array $options)
    {
    }
\
```

phpstan would say: 

```
$options type has no value type specified in iterable type array
```

its the same case as the FormTypeInterface Stub

ondrejmirtes:
How does the `$options` array going into these methods usually looks like?

pscheit-lillydoo:
e.g. for a collection type 

```php
 [
    'entry_type' => EmailType::class,
    'entry_options' => [
        'attr' => ['class' => 'email-box'],
    ],
]);
```

basically everything mentioned here https://symfony.com/doc/current/reference/forms/types.html#other-fields
listed as options of the form types.  
But could be custom options, too, depends on the formType (if it is a custom one, too)

ondrejmirtes:
So is it `array<string, mixed>`?

pscheit-lillydoo:
yes sir, exactly as stated in the stub

stubs/FormTypeExtensionInterface.stub
```diff
@@ -0,0 +1,21 @@
+<?php
+
+namespace Symfony\Component\Form;
+
+interface FormTypeExtensionInterface
+{
+    /**
+     * @param array<mixed> $options
```
pscheit-lillydoo, line 8: this is the correct type

ondrejmirtes:
Nope, your stub says `array<mixed>`.

pscheit-lillydoo:
arg, right 👍

ondrejmirtes:
Push the fix here in this PR, thanks.

ondrejmirtes:
Thank you, tagged as 0.12.10.

PR #110, opened 2020-10-29
(feat): add stubs for symfony/serializer interfaces

lucassabreu:


ondrejmirtes:
Hi, there aren't any generics in these stubs, what do they actually solve?

lucassabreu:
hi @ondrejmirtes, today there isn't any generics on these classes.

the pr is mostly to set a element type for the `$context` array on these classes, phpstan is requiring a type for it on higher levels. but in some situations these parameter is not used, so add these stubs would mean we don't have to set it on all classes.

we could say the `NormalizerInterface` could be use a generic type for the return of the `normalize` method, but it can set it in the implementation.

stubs/NormalizerInterface.stub
```diff
@@ -0,0 +1,14 @@
+<?php
+
+namespace Symfony\Component\Serializer\Normalizer;
+
+interface NormalizerInterface
+{
+    /**
+     * @param mixed $object
```
ondrejmirtes, line 8: could be `object`, right?

ondrejmirtes (changes requested):

ondrejmirtes:
The build failures need fixing :)

```
------ ------------------------------------------------------------------- 

  Line   stubs/ContextAwareDenormalizerInterface.stub                       

 ------ ------------------------------------------------------------------- 

         Reflection error:                                                  

         Symfony\Component\Serializer\Normalizer\DenormalizerInterface not  

         found.                                                             

 ------ ------------------------------------------------------------------- 

 ------ ----------------------------------------------------------------------- 

  Line   stubs/DenormalizableInterface.stub                                     

 ------ ----------------------------------------------------------------------- 

  15     Parameter $denormalizer of method                                      

         Symfony\Component\Serializer\Normalizer\DenormalizableInterface::deno  

         rmalize() has invalid typehint type                                    

         Symfony\Component\Serializer\Normalizer\DenormalizerInterface.         

  15     Parameter $denormalizer of method                                      

         Symfony\Component\Serializer\Normalizer\DenormalizableInterface::deno  

         rmalize() has invalid typehint type                                    

         Symfony\Component\Serializer\Normalizer\DenormalizerInterface.         

 ------ ----------------------------------------------------------------------- 
```

stubs/NormalizerInterface.stub
lucassabreu: it could be... but could be a stream, resource, scalar or other thing, the supports method limits what it can be... 

but i think most times will be object.

should i change to `mixed|object`?

lucassabreu: just to be sure, symfony's serializer does allow for non-objects to be "normalized", i never used it, but is a option.

https://github.com/symfony/serializer/blob/5.x/Serializer.php#L236

ondrejmirtes: `mixed|object` would be the same as `mixed`.

stubs/EncoderInterface.stub
```diff
@@ -0,0 +1,20 @@
+<?php
+
+namespace Symfony\Component\Serializer\Encoder;
+
+interface EncoderInterface
+{
+    /**
+     * @param mixed|object|array<mixed> $data
```
ondrejmirtes, line 8: This is same as `mixed`.

stubs/NormalizerInterface.stub
```diff
@@ -0,0 +1,14 @@
+<?php
+
+namespace Symfony\Component\Serializer\Normalizer;
+
+interface NormalizerInterface
+{
+    /**
+     * @param mixed|object|array<mixed> $object
```
ondrejmirtes, line 8: This is same as `mixed`.

ondrejmirtes (changes requested):
Alright, one last round and we can merge this :)

stubs/NormalizerInterface.stub
lucassabreu: fixed.

ondrejmirtes:
Thank you!

ondrejmirtes:
Tagged as 0.12.11.

ossinkine:
Hey @lucassabreu, why these stubs are needed? Symfony already has types in these interfaces and more accurate, for example `EncoderInterface::encode` returns only `string`. This produces false positives.

mhujer:
@ossinkine it seems that the types were fixed in Symfony (https://github.com/symfony/symfony/pull/39451) after this PR was merged.

ossinkine:
@mhujer Thank you but this does not explain why these stubs were created.

lucassabreu:
> Hey @lucassabreu, why these stubs are needed? Symfony already has types in these interfaces and more accurate, for example `EncoderInterface::encode` returns only `string`. This produces false positives.

Hi @ossinkine as mhujer said these interfaces didn't have the types for return and parameters, and we had to set them on every implementation before to satisfy higher levels on phpstan, but this types were ways the same and didn't make much sense to set them on my classes that were just implementing these interfaces.

the types on this pr are the ones I thought were right at the time.

if symfony has them set now i think that we can remove these stubs now for the new symfony versions.

PR #80, opened 2020-04-24
Add stubs for various classes

emodric:
This adds stubs for:

* `Symfony\Component\Form\FormTypeExtensionInterface`
* `Symfony\Component\Serializer\Normalizer\NormalizerInterface`
* `Symfony\Component\HttpFoundation\ParameterBag`
* `Symfony\Component\Form\DataMapperInterface`

I also added a stub for `Twig\Node\Node`, however, if you think this is not a place for it, I will remove it. In that case, do you know if there's an official repo I can contribute to?

PR #122, opened 2021-01-11
Add stubs for Twig Node and Symfony ParameterBag

emodric:
Repeat of #80 with already existing stubs removed.

These solve errors like:

```
Method AppBundle\Templating\Twig\TokenParser\MyNode::
parse() return type has no value type specified in iterable type Twig\Node\Node.
💡 Consider adding something like Twig\Node\Node<Foo> to the PHPDoc.
```

or

```
Method AppBundle\Controller\Test::validateRequestData()
has parameter $data with no value type specified in iterable type Symfony\Component\HttpFoundation\ParameterBag.
💡 Consider adding something like Symfony\Component\HttpFoundation\ParameterBag<Foo> to the PHPDoc.
```

### DIFF
--- a/extension.neon
+++ b/extension.neon
@@ -24,8 +24,10 @@ parameters:
 		- stubs/FormTypeInterface.stub
 		- stubs/FormView.stub
 		- stubs/HeaderBag.stub
+		- stubs/Node.stub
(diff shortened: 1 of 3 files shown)


ondrejmirtes:
Perfect, thank you!

PR #123, opened 2021-01-25
add stub for Cookie

rgrassian:
Hi @ondrejmirtes 

see: https://github.com/symfony/symfony/blob/5.x/src/Symfony/Component/HttpFoundation/Cookie.php#L247


ondrejmirtes (changes requested):
I'd say the `withSameSite()` method should also be covered. Thanks.

ondrejmirtes:
Thank you!

PR #124, opened 2021-01-25
Adds stub file for PropertyPathInterface

faizanakram99:


faizanakram99:
> X and Y were meant to be just placeholders, you should type what it's really iterating over.

yep, forgot to add @template :sweat_smile: 
Fixed

ondrejmirtes:
This will just cause another error when you typehint the interface. I suppose the iterated types are always the same, check out the implementations what's going on in them.

faizanakram99:
> This will just cause another error when you typehint the interface. I suppose the iterated types are always the same, check out the implementations what's going on in them.

You're right, it showed implementations should be contravariant with parent errors.
From the implementations it looks like key is always integer and value is always string, fixed.

Thank you

ondrejmirtes:
Thank you!

PR #126, opened 2021-02-03
Support for TreeBuilder::getRootNode()

ruudk:
After seeing #60 stuck I decided to give it a shot... 

This is very naive, and it assumes the `TreeBuilder` is always of type `array`.

```
TreeBuilder::__construct(string $name, string $type = 'array', NodeBuilder $builder = null)
```

If I am able to find the second `$type` parameter to the `new TreeBuilder()` call then I can make the return type of `getRootNode` dynamic based on that.

@ondrejmirtes Is this even possible?

The issue with NodeDefinition::end() is that it can return `null` when you've reached the root. The current extension removes the nullability in favor of DX. But it's also very naive. No idea if that can be done dynamically.

Hope to get some advice on how to proceed.... 🙏

### DIFF
--- a/src/Type/Symfony/TreeBuilderDynamicReturnTypeExtension.php
+++ b/src/Type/Symfony/TreeBuilderDynamicReturnTypeExtension.php
@@ -1,0 +1,68 @@
+<?php declare(strict_types = 1);
+
+namespace PHPStan\Type\Symfony;
+
+use PhpParser\Node\Expr\StaticCall;
+use PhpParser\Node\Name;
+use PHPStan\Analyser\Scope;
+use PHPStan\Broker\Broker;
+use PHPStan\Reflection\BrokerAwareExtension;
+use PHPStan\Reflection\MethodReflection;
+use PHPStan\Reflection\ParametersAcceptorSelector;
+use PHPStan\Type\ConstantScalarType;
+use PHPStan\Type\DynamicStaticMethodReturnTypeExtension;
+use PHPStan\Type\Type;
+
+final class TreeBuilderDynamicReturnTypeExtension implements DynamicStaticMethodReturnTypeExtension, BrokerAwareExtension
+{
+
+	/** @var Broker */
+	private $broker;
+
+	public function setBroker(Broker $broker): void
+	{
+		$this->broker = $broker;
+	}
+
+	public function getClass(): string
+	{
+		return 'Symfony\Component\Config\Definition\Builder\TreeBuilder';
+	}
+
+	public function isStaticMethodSupported(MethodReflection $methodReflection): bool
+	{
+		return $methodReflection->getName() === '__construct';
+	}
+
+	public function getTypeFromStaticMethodCall(MethodReflection $methodReflection, StaticCall $methodCall, Scope $scope): Type
+	{
+		if (!$methodCall->class instanceof Name) {
+			throw new \PHPStan\ShouldNotHappenException();
+		}
+
+		$className = $scope->resolveName($methodCall->class);
+		if (!$this->broker->hasClass($className)) {
+			return ParametersAcceptorSelector::selectSingle(
+				$methodReflection->getVariants()
+			)->getReturnType();
+		}
+
+		$args = [];
+		foreach ($methodCall->args as $arg) {
+			$value = $scope->getType($arg->value);
+
+			if (!$value instanceof ConstantScalarType) {
+				throw new \LogicException();
+			}
+
+			$args[] = $value->getValue();
+		}
+
+		$treeBuilder = new $className(
+			...$args
+		);
+
+		return new TreeBuilderType($className, $treeBuilder->getRootNode());
+	}
+
+}
--- a/src/Type/Symfony/TreeBuilderDynamicReturnTypeExtension.php
+++ b/src/Type/Symfony/TreeBuilderDynamicReturnTypeExtension.php
@@ -5,24 +5,30 @@ namespace PHPStan\Type\Symfony;
 use PhpParser\Node\Expr\StaticCall;
 use PhpParser\Node\Name;
 use PHPStan\Analyser\Scope;
-use PHPStan\Broker\Broker;
-use PHPStan\Reflection\BrokerAwareExtension;
 use PHPStan\Reflection\MethodReflection;
-use PHPStan\Reflection\ParametersAcceptorSelector;
-use PHPStan\Type\ConstantScalarType;
 use PHPStan\Type\DynamicStaticMethodReturnTypeExtension;
 use PHPStan\Type\Type;
+use PHPStan\Type\TypeUtils;
+use Symfony\Component\Config\Definition\Builder\ArrayNodeDefinition;
+use Symfony\Component\Config\Definition\Builder\BooleanNodeDefinition;
+use Symfony\Component\Config\Definition\Builder\EnumNodeDefinition;
+use Symfony\Component\Config\Definition\Builder\FloatNodeDefinition;
+use Symfony\Component\Config\Definition\Builder\IntegerNodeDefinition;
+use Symfony\Component\Config\Definition\Builder\ScalarNodeDefinition;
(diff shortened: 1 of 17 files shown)


ondrejmirtes:
@ruudk Hi, I need a sample of analyzed code and what is this extension trying to achieve.

ruudk:
@ondrejmirtes I'm trying to let PHPStan understand this: https://github.com/symfony/symfony/blob/60118f14aa33165407b06ed59225a612049f715f/src/Symfony/Bundle/DebugBundle/DependencyInjection/Configuration.php#L30-L55

If you want to reproduce it:

```
git clone https://github.com/symfony/symfony.git
cd symfony
composer install
composer req --dev phpstan/phpstan:^0.12.71
vendor/bin/phpstan analyse --level=8 src/Symfony/Bundle/DebugBundle/DependencyInjection/Configuration.php

 ------ -----------------------------------------------------------------------
  Line   Configuration.php
 ------ -----------------------------------------------------------------------
  33     Call to an undefined method
         Symfony\Component\Config\Definition\Builder\NodeDefinition::children(
         ).
  58     Call to an undefined method
         Symfony\Component\Config\Definition\Builder\NodeDefinition::children(
         ).
 ------ -----------------------------------------------------------------------


 [ERROR] Found 2 errors
```

src/Type/Symfony/TreeBuilderDynamicReturnTypeExtension.php
```diff
@@ -0,0 +1,33 @@
...
+		return $methodReflection->getName() === 'getRootNode';
+	}
+
+	public function getTypeFromMethodCall(
+		MethodReflection $methodReflection,
+		MethodCall $methodCall,
+		Scope $scope
+	): Type
+	{
+		return new \PHPStan\Type\ObjectType('Symfony\Component\Config\Definition\Builder\ArrayNodeDefinition');
```
ondrejmirtes, line 30: There's no logic at all, you can simply write a stub file (or send Symfony a PR that fixes the PHPDoc). https://phpstan.org/user-guide/stub-files

src/Type/Symfony/NodeDefinitionDynamicReturnTypeExtension.php
```diff
@@ -0,0 +1,33 @@
...
+		return $methodReflection->getName() === 'end';
+	}
+
+	public function getTypeFromMethodCall(
+		MethodReflection $methodReflection,
+		MethodCall $methodCall,
+		Scope $scope
+	): Type
+	{
+		return new \PHPStan\Type\ObjectType('Symfony\Component\Config\Definition\Builder\ArrayNodeDefinition');
```
ondrejmirtes, line 30: There's no logic at all, you can simply write a stub file (or send Symfony a PR that fixes the PHPDoc). https://phpstan.org/user-guide/stub-files

ondrejmirtes (changes requested):

src/Type/Symfony/NodeDefinitionDynamicReturnTypeExtension.php
ruudk: In the description I wrote why there is no logic yet: I want to know if I can somehow read the 2nd argument of the constrctor of the TreeBuilder. That one defines the type....So when the 2nd argument is `new TreeBuilder('name', 'array')` then `getRootNode()` returns `ArrayNodeDefinition`. 

There is a list of some string inputs to *NodeDefinition types... but it might be that this cannot be read from the source code...

ondrejmirtes: 1) You can solve it with generics although that would require `TreeBuilder` to be generic so that each place where it's typehinted would require to specify the type, like `TreeBuilder<array>`. Then you'd be able to read the type in the extension from `GenericObjectType`.
2) Or for local code, you can write a dynamic return type extension (`DynamicStaticMethodReturnTypeExtension`) for `TreeBuilder::__construct()` which gets invoked for `new` as well. And return a new `Type` class that extends `ObjectType` and carries the additional information, which you in turn read in `getRootNode` dynamic return type extension as well. Something like this is already done in `phpstan-doctrine` in `QueryBuilderType` (https://github.com/phpstan/phpstan-doctrine/blob/master/src/Type/Doctrine/QueryBuilder/QueryBuilderType.php).

ruudk: @ondrejmirtes Thanks for the suggestion. I went for option 2. It works 🎉  Amazing.

I decided to remove the `->end()` functionality. Let's first finish and merge this. Then I'll create a new PR with tests to work on supporting `->end()`.

ruudk:
@ondrejmirtes It's green ✅  🎉

src/Type/Symfony/TreeBuilderDynamicReturnTypeExtension.php
```diff
@@ -0,0 +1,68 @@
...
+use PHPStan\Reflection\MethodReflection;
+use PHPStan\Reflection\ParametersAcceptorSelector;
+use PHPStan\Type\ConstantScalarType;
+use PHPStan\Type\DynamicStaticMethodReturnTypeExtension;
+use PHPStan\Type\Type;
+
+final class TreeBuilderDynamicReturnTypeExtension implements DynamicStaticMethodReturnTypeExtension, BrokerAwareExtension
+{
+
+	/** @var Broker */
```
ondrejmirtes, line 19: Broker is obsolete. You can also remove `implements BrokerAwareExtension` and the setter and ask for `ReflectionProvider` in the constructor instead.

src/Type/Symfony/TreeBuilderDynamicReturnTypeExtension.php
```diff
@@ -0,0 +1,68 @@
...
+				$methodReflection->getVariants()
+			)->getReturnType();
+		}
+
+		$args = [];
+		foreach ($methodCall->args as $arg) {
+			$value = $scope->getType($arg->value);
+
+			if (!$value instanceof ConstantScalarType) {
+				throw new \LogicException();
```
ondrejmirtes, line 55: Extensions shouldn't throw any exception, you can easily cause "internal error" while analyzing code. Instead, you should return the default return type. See for example: https://github.com/phpstan/phpstan-symfony/blob/f051037bfd2d6567cb48a5f09a6ba8e78c82986e/src/Type/Symfony/InputInterfaceGetArgumentDynamicReturnTypeExtension.php#L43

src/Type/Symfony/TreeBuilderDynamicReturnTypeExtension.php
```diff
@@ -0,0 +1,68 @@
...
+			}
+
+			$args[] = $value->getValue();
+		}
+
+		$treeBuilder = new $className(
+			...$args
+		);
```
ondrejmirtes, line 63: We should take a more "static" approach. You can also easily cause an internal error when analyzing code, for example when the constructor call has wrong arguments.

We should have some hardcoded map of what the type => root node types are. And I guess the `TreeBuilderType` constructor 2nd argument should be `ObjectType`.

ondrejmirtes (changes requested):

src/Type/Symfony/TreeBuilderDynamicReturnTypeExtension.php
ruudk: Removed, it was not needed anyway.

src/Type/Symfony/TreeBuilderDynamicReturnTypeExtension.php
ruudk: Done.

src/Type/Symfony/TreeBuilderDynamicReturnTypeExtension.php
ruudk: I changed the code and I feel it's now much simpler. What do you think?

ondrejmirtes (approved):

ondrejmirtes:
Perfect, thank you!

PR #129, opened 2021-02-08
Fix phpstan callback type problem

michealmouner:
phpstan throw error for this stub
`vendor/phpstan/phpstan-symfony/stubs/ChoiceLoaderInterface.stub`

```
-
    message: "#^Method Symfony\\\\Component\\\\Form\\\\ChoiceList\\\\Loader\\\\ChoiceLoaderInterface\\:\\:loadChoicesForValues\\(\\) has parameter \\$value with no signature specified for callable\\.$#"
    count: 1
    path: vendor/phpstan/phpstan-symfony/stubs/ChoiceLoaderInterface.stub

-
    message: "#^Method Symfony\\\\Component\\\\Form\\\\ChoiceList\\\\Loader\\\\ChoiceLoaderInterface\\:\\:loadValuesForChoices\\(\\) has parameter \\$value with no signature specified for callable\\.$#"
    count: 1
    path: vendor/phpstan/phpstan-symfony/stubs/ChoiceLoaderInterface.stub
```

I updated annotation to fix this error for phpstan

related docs
https://phpstan.org/user-guide/stub-files

the above document said:
```
What about the analysis of the stub files?

Stub files don’t need to be added to the list of analysed paths. They’re analysed independently of the chosen rule level with a set of select rules mainly evaluating the PHPDoc well-formedness.

Errors reported in the stub files also can’t be ignored and all of them must be fixed.
```

### DIFF
--- a/stubs/ChoiceLoaderInterface.stub
+++ b/stubs/ChoiceLoaderInterface.stub
@@ -6,7 +6,7 @@ interface ChoiceLoaderInterface
 {
     /**
      * @param array<string> $values
-     * @param callable|null $value
+     * @param callable(mixed): void|null $value
      *
      * @return array<mixed>
      */
@@ -14,7 +14,7 @@ interface ChoiceLoaderInterface
 
     /**
      * @param array<mixed> $choices
-     * @param callable|null $value
+     * @param callable(mixed): void|null $value
      *
      * @return array<string>
      */


ondrejmirtes:
Hi, how does typically calling these methods look like? What's the parameter going into the callback?

michealmouner:
> Hi, how does typically calling these methods look like? What's the parameter going into the callback?

actually it depends on the case/use
for Doctrine choices with `SonataAdminBundle` .. the callback takes `Entity|null`
other/default usage I found it doesn't send any params to the callback

ondrejmirtes:
@michealmouner This update would actually cause an error. The PHPDoc is saying `mixed can be passed to the callback`, but if user decides to have something more specific in the parameter (like an object), PHPStan would error.

michealmouner:
> @michealmouner This update would actually cause an error. The PHPDoc is saying `mixed can be passed to the callback`, but if user decides to have something more specific in the parameter (like an object), PHPStan would error.

I think it will not cause error (maybe I missed something) .. what is your solution for these error?

ondrejmirtes:
I think it's best to leave it at `callable`. I just commited a change that will disable this for the stubs (https://github.com/phpstan/phpstan-src/commit/b7d60809dd3c1fed00dde616df6c6946c7da118d) even if the project's configuration has it enabled.

Issue #130, opened 2021-02-11
Call to an undefined method Symfony\Component\Config\Definition\Builder\NodeParentInterface::scalarNode()


bendavies:
Hi there!

since updating to 0.12.17, I'm now receiving:
```
Call to an undefined method Symfony\Component\Config\Definition\Builder\NodeParentInterface::scalarNode().
``` 
for code such as:

```php
$rootNode = $treeBuilder->getRootNode();
$rootNode
    ->children()
    ->scalarNode('protocol')->end()
    ->scalarNode('host')->end()
    ->scalarNode('resource')->end()
    ->end()
;
```

Thanks!

ruudk:
Working on that today!

PR #131, opened 2021-02-12
Add full support for Config/TreeBuilder 

ruudk:
Fixes https://github.com/phpstan/phpstan-symfony/issues/130

It took a bit more work than I expected. 

I have tested this on my project and it works fine now.


ruudk:
@bendavies Could you please validate if it works?

Apply this to your `composer.json`:
```json
    "repositories": [
        {
            "type": "vcs",
            "url": "http://github.com/ruudk/phpstan-symfony"
        }
    ],
    "require-dev": {
        "phpstan/phpstan-symfony": "dev-full-tree-builder"
    }
```
and then run `composer update phpstan/phpstan-symfony`

bendavies:
seems to have fixed it yes!

src/Type/Symfony/Config/NodeDefinitionEndDynamicReturnTypeExtension.php
```diff
@@ -0,0 +1,39 @@
...
+namespace PHPStan\Type\Symfony\Config;
+
+use PhpParser\Node\Expr\MethodCall;
+use PHPStan\Analyser\Scope;
+use PHPStan\Reflection\MethodReflection;
+use PHPStan\Reflection\ParametersAcceptorSelector;
+use PHPStan\Type\DynamicMethodReturnTypeExtension;
+use PHPStan\Type\Type;
+
+final class NodeDefinitionEndDynamicReturnTypeExtension implements DynamicMethodReturnTypeExtension
```
ondrejmirtes, line 12: This class looks exactly the same as NodeBuilderEndDynamicReturnTypeExtension . Instead of duplicating the whole class, you can have only a single class, but making it dynamic by asking for $class in the constructor, then registering the class as two services in extension.neon, with different arguments.

ondrejmirtes (changes requested):

src/Type/Symfony/Config/NodeDefinitionEndDynamicReturnTypeExtension.php
ruudk: Much better indeed.

ruudk:
@ondrejmirtes Applied feedback ✅

ruudk:
I think it can be done even simpler.... let me give it a shot.

ruudk:
Tested it on my project again, all good. 

@bendavies If you can, please give it another run

extension.neon
```diff
@@ -146,12 +146,47 @@ services:
...
 
+	# ArrayNodeDefinition::*prototype() return type
+	-
+		factory: PHPStan\Type\Symfony\Config\ArrayNodeDefinitionPrototypeDynamicReturnTypeExtension
+		tags: [phpstan.broker.dynamicMethodReturnTypeExtension]
+
+	# ExprBuilder::end() return type
+	-
+		factory: PHPStan\Type\Symfony\Config\ReturnParentDynamicReturnTypeExtension
+		arguments: ['Symfony\Component\Config\Definition\Builder\ExprBuilder', ['end']]
```
ondrejmirtes, line 157: What I like is this style of arguments (where the constructor parameter name is mentioned in the config): https://github.com/phpstan/phpstan-src/blob/990ba51cb38f29b17579c54c1967223fcc56661a/conf/config.neon#L327-L332

ondrejmirtes (changes requested):
Otherwise 👍

extension.neon
ruudk: Better indeed, applied and ✅

ondrejmirtes (approved):

ondrejmirtes:
Alright, let's test this in the wild. I'm gonna tag a release right away.

ondrejmirtes:
Thank you!

PR #132, opened 2021-02-15
Fix beforeNormalization

VincentLanglet:
Related to https://github.com/phpstan/phpstan-symfony/pull/131 cc @ruudk 

BeforeNormalization has the same behavior than validate.

This fix an issue with 
```
->children()
    ->arrayNode('information')
        ->useAttributeAsKey('id')
        ->prototype('array')
            ->beforeNormalization()
                ->ifString()
                ->then(static function ($v) {
                    return [$v];
                })
            ->end()
            ->prototype('scalar')->end()
        ->end(
```
```
Call to an undefined method Symfony\Component\Config\Definition\Builder\NodeDefinition::prototype().
```

I tried it on my project and it fix the error. A release after the merge could be great if you have time @ondrejmirtes :)

### DIFF
--- a/tests/Type/Symfony/Config/TreeBuilderTest.php
+++ b/tests/Type/Symfony/Config/TreeBuilderTest.php
@@ -4,7 +4,6 @@ namespace PHPStan\Type\Symfony\Config;
 
 use Iterator;
 use PHPStan\Type\Symfony\ExtensionTestCase;
-use Symfony\Component\Config\Definition\Builder\TreeBuilder;
 
 final class TreeBuilderTest extends ExtensionTestCase
 {
--- a/tests/Type/Symfony/Config/TreeBuilderTest.php
+++ b/tests/Type/Symfony/Config/TreeBuilderTest.php
@@ -138,50 +138,50 @@ final class TreeBuilderTest extends ExtensionTestCase
 		yield ['
 		$arrayRootNode
 			->children()
-                ->arrayNode("methods")
-                    ->prototype("scalar")
-                        ->defaultNull()
-                    ->end()
-                ->end()
-            ->end()
+				->arrayNode("methods")
+					->prototype("scalar")
+						->defaultNull()
+					->end()
+				->end()
+			->end()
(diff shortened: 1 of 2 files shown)


ruudk:
Could you please also add a test case in https://github.com/phpstan/phpstan-symfony/blob/cc9d9ac264b8ac04961a03dc5b481b58ceaac248/tests/Type/Symfony/Config/TreeBuilderTest.php#L173-L185  ?

tests/Type/Symfony/Config/TreeBuilderTest.php
```diff
@@ -14,20 +14,6 @@ final class TreeBuilderTest extends ExtensionTestCase
...
-			->arrayNode('methods')
-			->prototype('scalar')
-			->validate()
-			->ifNotInArray(['one', 'two'])
-			->thenInvalid('%s is not a valid method.')
-			->end()
-			->end()
-			->end()
-			->end();
-
```
ruudk, line 30: Great. I forgot to remove these.

VincentLanglet:
I added the test

tests/Type/Symfony/Config/TreeBuilderTest.php
```diff
@@ -183,6 +168,21 @@ public function getProvider(): Iterator
...
+                    ->prototype("array")
+                        ->beforeNormalization()
+                            ->ifString()
+                            ->then(static function ($v) {
+								return [$v];
+							})
+                        ->end()
+                    ->end()
+                ->end()
+            ->end()
```
ruudk, line 184: While you're at it, could you please fix the indentation of this test case (plus the ones above that I forgot too)? 🙏 Would make it easier to understand.

OskarStark (approved):

tests/Type/Symfony/Config/TreeBuilderTest.php
VincentLanglet: Done

ruudk (approved):
Great! 👏

ondrejmirtes:
Thank you!

PR #139, opened 2021-03-08
Configure dynamic constant Kernel::VERSION_ID

ruudk:
Or should this be done by everyone on their own?


ondrejmirtes:
Thank you!

PR #144, opened 2021-04-13
Add support for InputBag::all (Symfony 5+)

leongersen:


src/Type/Symfony/InputBagDynamicReturnTypeExtension.php
```diff
@@ -21,14 +25,29 @@ public function getClass(): string
...
+		return in_array($methodReflection->getName(), ['get', 'all'], true);
 	}
 
 	public function getTypeFromMethodCall(
 		MethodReflection $methodReflection,
 		MethodCall $methodCall,
 		Scope $scope
 	): Type
+	{
+		switch ($methodReflection->getName()) {
```
ondrejmirtes, line 37: I don't like switch for these scenarios, please use a simple `if` statement.

It can also be a totally separate extension as these methods don't have anything in common.

ondrejmirtes (changes requested):

src/Type/Symfony/InputBagDynamicReturnTypeExtension.php
leongersen: I've updated the PR. I took the pattern from `ServiceDynamicReturnTypeExtension`.

ondrejmirtes (approved):

ondrejmirtes:
Thank you.

PR #145, opened 2021-04-13
Set php line endings for windows users with core.autocrlf enabled

leongersen:
This is [configured](https://github.com/phpstan/phpstan-src/blob/master/.gitattributes) for the main PhpStan source as well.


ondrejmirtes:
Thank you.

Issue #141, opened 2021-04-11
console-application.php for Symfony 5.1?


ThomasLandauer:
Since `config/bootstrap.php` doesn't exist anymore in Symfony 5.1 (see https://github.com/symfony/symfony-docs/issues/13759), the example `console-application.php` at https://github.com/phpstan/phpstan-symfony#console-command-analysis doesn't work anymore.

I did it quick and dirty like so, but what's the recommended way (to also be shown in README)?:
```php
use App\Kernel;
$kernel = new Kernel('dev', false);
return new \Symfony\Bundle\FrameworkBundle\Console\Application($kernel);
```

`object-manager.php` at https://github.com/phpstan/phpstan-doctrine#configuration is also affected.

mhujer:
My `console-boostrap.php` looks like this:

```php
<?php declare(strict_types = 1);

use App\Kernel;
use Symfony\Bundle\FrameworkBundle\Console\Application;
use Symfony\Component\Dotenv\Dotenv;

require dirname(__DIR__) . '/../vendor/autoload.php';

(new Dotenv())->bootEnv(dirname(__DIR__) . '/../.env');

$kernel = new Kernel($_SERVER['APP_ENV'], (bool) $_SERVER['APP_DEBUG']);
return new Application($kernel);
```

And `doctrine-orm-bootstrap.php`:

```php
<?php declare(strict_types = 1);

use App\Kernel;
use Symfony\Component\Dotenv\Dotenv;

require dirname(__DIR__) . '/../vendor/autoload.php';

(new Dotenv())->bootEnv(dirname(__DIR__) . '/../.env');

$kernel = new Kernel($_SERVER['APP_ENV'], (bool) $_SERVER['APP_DEBUG']);
$kernel->boot();
return $kernel->getContainer()->get('doctrine')->getManager();
```

ondrejmirtes:
@mhujer Can you please send a PR here and in phpstan-doctrine that updates the instructions in the README? I hope the new way is also backward compatible with older Symfony/Doctrine versions. Thank you.

ThomasLandauer:
For Symfony 4.4 they're recommending to have a `bootstrap.php`:
https://symfony.com/doc/4.4/configuration/dot-env-changes.html#updating-my-application
vs.
https://symfony.com/doc/current/configuration/dot-env-changes.html#updating-my-application

PR #150, opened 2021-04-23
Adding console-application.php for Symfony 5

ThomasLandauer:
* The code is taken from https://github.com/phpstan/phpstan-symfony/issues/141#issuecomment-817485623
* In the introduction I emphasized more that this is only relevant if you have Console Commands (right?)

Closes https://github.com/phpstan/phpstan-symfony/issues/141

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,29 +83,47 @@ parameters:
 
 Be aware that it may hide genuine errors in your application.
 
-## Console command analysis
+## Analysis of Symfony Console Commands
 
-You can opt in for more advanced analysis by providing the console application from your own application. This will allow the correct argument and option types to be inferred when accessing `$input->getArgument()` or `$input->getOption()`.
+You can opt in for more advanced analysis of [Symfony Console Commands](https://symfony.com/doc/current/console.html)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -83,30 +83,48 @@ parameters:
...
+
+use App\Kernel;
+use Symfony\Bundle\FrameworkBundle\Console\Application;
+use Symfony\Component\Dotenv\Dotenv;
 
-```bash
-Compile Error: Cannot Declare interface PhpParser\NodeVisitor, because the name is already in use
+require __DIR__ . '/../vendor/autoload.php';
+
+(new Dotenv())->bootEnv(__DIR__ . '/../.env');
```
ondrejmirtes, line 118: Could you unify the two scripts more? Either have `use` statements in everything for all names, or FQN for everything... Thanks!

ondrejmirtes (changes requested):

README.md
ThomasLandauer: Sure :-)

ondrejmirtes:
Thank you!

PR #151, opened 2021-04-26
Support for ServiceLocator

andrey-bondar:
Now when private services are fetching from Psr\Container\ContainerInterface::get(), an error "Service "%s" is private" occurred.

But Symfony has [ServiceLocators](https://symfony.com/doc/current/service_container/service_subscribers_locators.html#defining-a-service-locator) in order to define smaller containers with public access to services.

This PR solves error "Service "%s" is private" when service is fetching from ServiceLocator.


ondrejmirtes:
Hi, what errors it resolves? Please attach a test too.

andrey-bondar:
@ondrejmirtes I've attached tests

ondrejmirtes:
Thank you!

Wirone:
Sorry for digging the old issue @ondrejmirtes but I'm implementing Symfony extension in our app and I got "Service ... is private" errors. I've found out that our app uses service locators BUT they are not extending `Symfony\Component\DependencyInjection\ServiceLocator` but only implementing `Psr\Container\ContainerInterface`. I can't extend Symfony's `ServiceLocator` because we already have custom implementation of `get()`, so all I could do is implementing `Symfony\Contracts\Service\ServiceProviderInterface` (which is also implemented in `ServiceLocator`). Unfortunately it won't be resolved properly in `PHPStan\Rules\Symfony\ContainerInterfacePrivateServiceRule::processNode()` since it checks explicitly for `ServiceLocator`.

What do you think, is it possible to change condition (`$isServiceLocator` → `$isServiceProvider`) so raw `ServiceProviderInterface` would be supported instead of concrete `ServiceLocator`?

ondrejmirtes:
Feel free to send a PR with a test and a fix and we'll see (mostly what @lookyman will say about it :))

Issue #127, opened 2021-02-04
Support for ContainerInterface::getParameter return type


garak:
Problem: [ContainerInterface](https://github.com/symfony/symfony/blob/5.x/src/Symfony/Component/DependencyInjection/ContainerInterface.php) returns `array|bool|float|int|string|null` as value.

This is causing false positive in phpstan analysis, when you know that returned value has a specific type and you try to use it. For example, if you retrieve a string using `$container->getParameter('my_parameter')`, you currently can't do a concatenation, since phpstan is complaining about `Binary operation "." [...] results in an error`

It would be nice if the plugin could be able to retrieve actual type from container and override the return type of `getParameter` method.

ondrejmirtes:
Hi, there's already an extension for retrieving services out of the container here https://github.com/phpstan/phpstan-symfony/blob/master/src/Type/Symfony/ServiceDynamicReturnTypeExtension.php, feel free to get inspired and implement it for parameters too.

FerJEP:
@garak Hi, Did you work around this issue?

garak:
> @garak Hi, Did you work around this issue?

not yet

Kocal:
Hi guys!

I'm actually working on it since I faced the issue some days ago.

I copy/pasted services-related code but for parameters and was able to build a `DefaultParameterMap` given the XML Symfony container.
Now I just need to create a phpstan extension and I will be ready to open a PR.

ondrejmirtes:
Implemented by https://github.com/phpstan/phpstan-symfony/pull/152, thanks @Kocal!

ondrejmirtes:
Enjoy it in phpstan-symfony 0.12.29 :)

PR #152, opened 2021-05-01
Dynamically type getters and hassers for parameters

Kocal:
Hi!
This is a proposal for #127.

I've re-used the same logic for services:
- read and parse the Symfony container XML dump (it would have been simpler with a PHP dump 😛)
- create a `ParameterMap` which contains all parameters

Dynamic return types have been implemented for:
- `ParameterBagInterface#get` and `ParameterBagInterface#has`
- `ContainerInterface#getParameter` and `ContainerInterface#hasParameter` 
- `AbstractController#getParameter` and `Controller#getParameter` 

~However, I'm not able to do the same for `AbstractController#getParameter` and `Controller#getParameter` when using `$this->getParameter('...')` inside a `AbstractController` and `Controller`. It always use the [return type from `AbstractController#getParameter`](https://github.com/symfony/symfony/blob/5.x/src/Symfony/Bundle/FrameworkBundle/Controller/AbstractController.php#L78). Is this a limitation from PHPStan or I did something wrong @ondrejmirtes?~

~Also in Symfony 4.0 (lowest deps), there is no method `AbstractController#getParameter()` and it make the tests failing. What is the best way to support lowest and highest Symfony version in tests and extension? What about using Symfony 4.4 as minimal requirements and dropping `Controller` usage? This is the only 4.x supported version right now.~
**EDIT:** Nevermind I fixed in c2d3f798e84fc9949f60ff28f45f09dc9d6cebe4 by using 4.4 as lowest version for `symfony/framework-bundle`, and manually create the default type if a parameter does not exist.


tests/Type/Symfony/ExtensionTestWithoutContainer.php
```diff
@@ -10,7 +10,15 @@ class ExtensionTestWithoutContainer extends TypeInferenceTestCase
...
-		yield from $this->gatherAssertTypes(__DIR__ . '/data/ExampleControllerWithoutContainer.php');
+		if (class_exists('Symfony\Bundle\FrameworkBundle\Controller\Controller')) {
+			yield from $this->gatherAssertTypes(__DIR__ . '/data/ExampleController.php');
+		}
+
+		if (!class_exists('Symfony\Bundle\FrameworkBundle\Controller\AbstractController')) {
+			return;
+		}
+
+		yield from $this->gatherAssertTypes(__DIR__ . '/data/ExampleAbstractController.php');
```
ondrejmirtes, line 21: I know that PHPCS forced you to do this but I'd prefer structure that allowed adding new files easily. So I'd suggest adding another dataProvider with a single `class_exists` condition and `yield from` with just a single file. And adding that provider to `testFileAsserts`.

ondrejmirtes (changes requested):
Hi, I love this! This is well thought out and I found only a single minor problem :) I wish that more people like you contributed to PHPStan :) Thank you.

tests/Type/Symfony/ExtensionTestWithoutContainer.php
ondrejmirtes: Actually I might resolve this myself because I really want to get this released 👍

ondrejmirtes:
Alright: https://github.com/phpstan/phpstan-symfony/pull/152/commits/a586e2495c2622d5494999e85c77ddd5573efaad

I'm gonna squash this and release it :)

ondrejmirtes (approved):

ondrejmirtes:
Thank you!

mhujer:
@Kocal 👍 👏  Just tested your branch on our project and it found the following:
(all of them are valid reports which need to be fixed in our code)

- `Offset 0 does not exist on array<string>|string|null.`
- `Cannot cast array<string>|string|null to int.` (at `(int) $rawAmount`)
- ` Parameter #1 $input of method FooBar::fromString() expects string|null, array<string>|string|null given.` (when passing a `$request->query->all()['amount']`)
- `Parameter #1 $parameters of method BarBaz::filterParameters() expects array<string>, array<string, array<string>|string> given` when passing `$request->query->all()`

ondrejmirtes:
Enjoy it in phpstan-symfony 0.12.29 :)

Kocal:
Nice! :D

PR #159, opened 2021-05-05
Add default phpunit testsuite

shyim:
No need to give the folder always in the Command

Still works: `./vendor/bin/phpunit -c tests tests`
Now works also `./vendor/bin/phpunit -c tests`
or better `composer test`

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -61,5 +61,8 @@
     ]
   },
   "minimum-stability": "dev",
-  "prefer-stable": true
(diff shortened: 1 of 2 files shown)


ondrejmirtes:
Sorry, I don't want to have this not unified. The new modern way is to have `phpunit.xml` in the root + Makefile instead of Phing. See https://github.com/phpstan/phpstan-src.

PR #163, opened 2021-05-17
Dynamically load XML container via PHP

### LINKED ISSUE #105: Support APP_ENV for determining container filename
I saw https://github.com/phpstan/phpstan-symfony/issues/61 and that works as an okay workaround, but I would posit that developers shouldn't have to carry the burden of dynamically generating a configuration file or symlinking a to a specific file when an extension-level solution could just as easily be introduced in the code. 

With that said, my suggestion would be to look for `APP_ENV` in the environment, and if it exists, use that to determine the `container_xml_path` automatically. Then we developers can omit that configuration option if we wanted to, or if we absolutely had to, we could still provide it and it would override automatically determining the value (i.e., use current logic). 

In other words, if I - the developer - omit `container_xml_path`, and provide `APP_ENV=test`, then this extension would produce `/path/to/my/var/cache/test/Project_KernelTestContainer.xml` file. But, if I omit `APP_ENV=test` or I provide `container_xml_path`, then the extension would use the current logic.

The use case for this is to be able to run phpstan in both development and CI environments.

ruudk:
This allows to automatically resolve the XML container configuration based on environment variables.

It also has the benefit that it will automatically create the XML container file when it's not present (as it boots the container once).

Fixes https://github.com/phpstan/phpstan-symfony/issues/105

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,6 +68,32 @@ parameters:
         container_xml_path: var/cache/dev/App_KernelDevDebugContainer.xml
 ```
 
+The XML file should exist prior to running PHPStan.
+
+If you want this file to be created automatically and resolved based on `APP_ENV` you can use this:
+
+```yaml
+parameters:
+    symfony:
+        container_xml_path: tests/container-loader.php
+```
+
+```php
+//tests/container-loader.php
+
+use App\Kernel;
+use Symfony\Component\Dotenv\Dotenv;
+
+require __DIR__ . '/../vendor/autoload.php';
+
+(new Dotenv())->bootEnv(__DIR__ . '/../.env');
+
+$kernel = new Kernel($_SERVER['APP_ENV'], (bool) $_SERVER['APP_DEBUG']);
+$kernel->boot();
+
+return file_get_contents($kernel->getContainer()->getParameter('debug.container.dump'));
+```
+
 ## Constant hassers
 
 Sometimes, when you are dealing with optional dependencies, the `::has()` methods can cause problems. For example, the following construct would complain that the condition is always either on or off, depending on whether you have the dependency for `service` installed:
--- a/extension.neon
+++ b/extension.neon
@@ -59,17 +59,23 @@ services:
 		arguments:
 		    consoleApplicationLoader: %symfony.console_application_loader%
 
+	# xml container resolver
+	-
+		factory: PHPStan\Symfony\XmlContainerResolver
+		arguments:
+		    containerXmlPath: %symfony.container_xml_path%
+
 	# service map
 	symfony.serviceMapFactory:
 		class: PHPStan\Symfony\ServiceMapFactory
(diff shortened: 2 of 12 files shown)


README.md
```diff
@@ -68,6 +68,32 @@ parameters:
...
 ```
 
+The XML file should exist prior to running PHPStan.
+
+If you want this file to be created automatically and resolved based on `APP_ENV` you can use this:
+
+```yaml
+parameters:
+    symfony:
+        container_xml_path: tests/container-loader.php
```
ondrejmirtes, line 78: This is a bit too magic for my taste. It's called `container_xml_path`. There should be a different config key for this scenario.

ondrejmirtes: If both config keys are filled  with a value, the extension should throw an exception so that the user knows they should provide either XML or PHP, but not both.

ondrejmirtes (changes requested):

README.md
ruudk: Hmm, not sure if that makes it more clear. 

The `container_xml_path` is just points to a PHP file that returns the XML. 

How would you call it?

ondrejmirtes: Maybe we should make it more general and instead have a file that returns the whole container (which we can turn into the XML ourselves for our own purposes).

ondrejmirtes: And then it could be called `container_loader`

ruudk: Before doing it like this, I tried to create a new `ServiceContainerServiceMapFactory` and `ServiceContainerParameterMapFactory` that would actually load the real container via a separate `container_loader` file.

The problem is, the container is compiled, and does not contain the data we need.

ondrejmirtes:
Looking at the issue I don't think we really need this complexity in the extension itself. The user can always dump the XML container based on whatever logic they want, to any path they want, and then run PHPStan. So you can do `make phpstan` which dumps the container with the right env, and then runs `vendor/bin/phpstan`.

ruudk:
I can create a script for sure. But it also means that the integration in PHPStorm will not work out of the box if you have no cache. Personally I `rm -rf var/cache` often. And I think more Symfony users do.

The console loader automatically boots and loads the application without manual work. In the Doctrine extension, the ObjectManager is automatically loaded without manual work.

Only this Service Container integration relies on a file to be present on disk and does not automatically creates it.

ondrejmirtes:
We could support container_loader that returns `$kernel->getContainer()`. The question is - what the user needs to do so that `$container->getParameter('debug.container.dump')` will always contain what we need?

b2p-fred:
Any news about this PR since May ?

ruudk:
I lost interest in solving this as this time. Feel free to pick it up from here. Maybe I'll come back to it in the future and we can reopen this PR.

raziel057:
I also encounter the following issue. The auto detection of XML container based on APP_ENV seems a good idea.

PR #171, opened 2021-06-02
Fix DenormalizableInterface::denormalize() return type declaration

jeroennoten:
See https://github.com/symfony/symfony/pull/39561

DenormalizableInterface::denormalize() should not return anything, so the `@return` declaration should be removed. If you look at the usage in `CustomNormalizer::denormalize()`, it becomes clear that this method shouldn't return anything:
```php
    public function denormalize($data, $type, $format = null, array $context = [])
    {
        $object = $this->extractObjectToPopulate($type, $context) ?: new $type();
        $object->denormalize($this->serializer, $data, $format, $context);

        return $object;
    }
```

Note that in this case `$object implements DenormalizableInterface`, so this is about the `$object->normalize()` call above.


ondrejmirtes:
Thank you.

PR #174, opened 2021-06-07
Allow boolean in Command->addOption default value (fixes #172)

### LINKED ISSUE #172: Command addOption fails with boolean default
Using a `bool` as default value for a command option with `InputOption::VALUE_REQUIRED` fails with `Parameter #5 $default of method Symfony\Component\Console\Command\Command::addOption() expects int|string|null, bool given.`.

The phpdoc for this function had
`@param string|string[]|int|bool|null $default     The default value (must be null for InputOption::VALUE_NONE)` until recently and current symfony versions seem to be using `string|string[]|bool|null` (i.e. removed `int`).
so this should be valid, but the relevant rule only includes `int|string|null`:
https://github.com/phpstan/phpstan-symfony/blob/0cd5b0d7666294c2a983aa276bb2d7fa125d0ae4/src/Rules/Symfony/InvalidOptionDefaultValueRule.php#L73 

Can bool be added to this rule? Alternatively, if there is a reason why bool was not included in that type could the error message be changed to explain it better as seeing this error which does not match the relevant phpdoc is very confusing.

codeTom:
The phpdoc in addOption() allows boolean as default value but the rule did not, this adds BooleanType to supported values and fixes #172


ondrejmirtes:
Hi, this change needs to be accompanied with a test. Should be easy: https://github.com/phpstan/phpstan-symfony/blob/master/tests/Rules/Symfony/InvalidOptionDefaultValueRuleTest.php

codeTom:
Hi, thanks for the response, I added a line that would fail before and should work now to ExampleCommand.php. The change shouldn't introduce new errors so I think that should work.

ondrejmirtes:
The build needs fixing.

codeTom:
Should be fixed now.

ondrejmirtes:
Thank you!

Issue #86, opened 2020-06-08
Call to method Request::hasSession() will always evaluate to true.


mmarton:
I'm not entirely sure this is a phpstan related issue.

When I check for session, ie:
```php
 if ($request->hasSession()) {
     $request->getSession()->set('key', 'value');
}
```

I got the folllowing phpstan error:
> Call to method Symfony\Component\HttpFoundation\Request::hasSession() will always evaluate to true.

If I comment out the use of phpstan-symfony OR the phpstan-strict-rules plugin, the error disappears.

Maybe related to https://github.com/phpstan/phpstan-symfony/issues/56

Symfony version 5.0
Phpstan versions:
phpstan/phpdoc-parser 0.4.4
phpstan/phpstan 0.12.26
phpstan/phpstan-deprecation-rules 0.12.4
phpstan/phpstan-doctrine 0.12.14
phpstan/phpstan-strict-rules 0.12.2
phpstan/phpstan-symfony 0.12.6

9ae8sdf76:
Same here. This is because Symfony defines the type as `SessionInterface` without it possibly being null. How can we work around this without disabling the remainder of the rules?

ondrejmirtes:
> This is because Symfony defines the type as SessionInterface without it possibly being null. 

This seems right. It's because of this extension: https://github.com/phpstan/phpstan-symfony/blob/9cd6919b8c695206dc24b5b5e8a082cd3b8f4c05/src/Type/Symfony/RequestTypeSpecifyingExtension.php

I think it was created back when `getSession()` returned nullable type in Symfony 4.x.

The extension should be modified to return empty `new SpecifiedTypes()` when the type is already non-nullable.

VincentLanglet:
> > This is because Symfony defines the type as SessionInterface without it possibly being null.
> 
> This seems right. It's because of this extension: https://github.com/phpstan/phpstan-symfony/blob/9cd6919b8c695206dc24b5b5e8a082cd3b8f4c05/src/Type/Symfony/RequestTypeSpecifyingExtension.php
> 
> I think it was created back when `getSession()` returned nullable type in Symfony 4.x.
> 
> The extension should be modified to return empty `new SpecifiedTypes()` when the type is already non-nullable.

In my project, I'm using Symfony 4.x ; so `getSession` is nullable.

And I'm curious how the `getSession` type specifier has an impact on the `hasSession` resolution

ondrejmirtes:
[...]

PR #181, opened 2021-06-30
Fix hasSession specifier

VincentLanglet:
Closes https://github.com/phpstan/phpstan-symfony/issues/86

I didn't succeed to reproduce the test proposed here https://github.com/phpstan/phpstan-symfony/issues/86#issuecomment-871350246

But the code I added were failing prior to this change.

### DIFF
--- a/src/Type/Symfony/RequestTypeSpecifyingExtension.php
+++ b/src/Type/Symfony/RequestTypeSpecifyingExtension.php
@@ -46,10 +46,15 @@ final class RequestTypeSpecifyingExtension implements MethodTypeSpecifyingExtens
 	{
 		$classReflection = $this->broker->getClass(self::REQUEST_CLASS);
 		$methodVariants = $classReflection->getNativeMethod(self::GET_METHOD_NAME)->getVariants();
+		$returnType = ParametersAcceptorSelector::selectSingle($methodVariants)->getReturnType();
+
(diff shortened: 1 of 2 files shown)


ondrejmirtes:
Please add your efforts reproducing this case, thanks.

VincentLanglet:
> Please add your efforts reproducing this case, thanks.

I had basically nothing.

My issue is that https://github.com/phpstan/phpstan-beberlei-assert/blob/master/tests/Type/BeberleiAssert/ImpossibleCheckTypeMethodCallRuleTest.php is a RuleTestCase but here, I'm testing a TypeSpecifier, so I don't have any rule to provide (and instead dunno how to provide the type specifier).

ondrejmirtes:
It works exactly the same way, you just need to provide a file that shows the code that fails (but shouldn't fail).

VincentLanglet:
> It works exactly the same way, you just need to provide a file that shows the code that fails (but shouldn't fail).

But what do I pass to
```
protected function getRule(): Rule
```
?

ondrejmirtes:
The same thing I linked, you just copy that part.

VincentLanglet:
> The same thing I linked, you just copy that part.

Oh my bad, I've updated the tests now. 
I check that `request_get_session.php` has no phpstan error. Without the fix I tried, and I had the "always true" error.

ondrejmirtes:
Thank you!

PR #184, opened 2021-07-16
Add back AclInterface stub

VincentLanglet:
HI @ondrejmirtes, the security stubs were fixing two things:
- The `AclProviderInterface`/`MutableAclProviderInterface` issue. This will be fixed in Symfony 5.4 and wasn't easily fixed in phpstan.
- The array keys of the `AclInterface` methods wich is int (and not string|int). I opened a PR on Symfony, but having the stub here allow to fix the issue for all the version of symfony (even < 5.4). For instance the sonata build is currently failing because of the removal of this stub.

So I think we should add it back.


ondrejmirtes:
Thank you!

PR #192, opened 2021-08-17
Ignore ContainerBagInterface in ContainerInterfaceUnknownServiceRule

### LINKED ISSUE #191: False positive with \Symfony\Component\DependencyInjection\ParameterBag\ContainerBagInterface::get()
Symfony `ContainerBagInterface` implements `Psr\Container\ContainerInterface` but actually [it returns a container parameter][1], not a service. 

However `ContainerInterfaceUnknownServiceRule` checks for service existence, not parameter. For example this code

```php
/** @var \Symfony\Component\DependencyInjection\ParameterBag\ContainerBagInterface $params */
$parameterValue = $params->get('parameter_name');
```

results in the following error

```
Service "parameter_name" is not registered in the container.
```

I suggest to ignore subtypes of `ContainerBagInterface` in the `ContainerInterfaceUnknownServiceRule`. Is this correct?

[1]:https://symfony.com/doc/current/configuration.html#accessing-configuration-parameters

alexsakhnov:
Fixes #191

I also added a test (not sure if it's correct)

### DIFF
--- a/src/Rules/Symfony/ContainerInterfaceUnknownServiceRule.php
+++ b/src/Rules/Symfony/ContainerInterfaceUnknownServiceRule.php
@@ -55,6 +55,11 @@ final class ContainerInterfaceUnknownServiceRule implements Rule
 		}
 
 		$argType = $scope->getType($node->var);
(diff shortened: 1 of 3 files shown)


ondrejmirtes:
> I also added a test (not sure if it's correct)

If it failed before changing `src/` then it's correct :)

alexsakhnov:
Thanks)

PR #193, opened 2021-08-23
Add Constraint::$errorNames

enumag:
Getting this error on a custom constraint. Adding the type in the stub fixes it.

```
Library/Symfony/Constraint/BankAccount.php:26
 ↳ Type array<string, string> of property App\Library\Symfony\Constraint\BankAccount::$errorNames is not the same as type mixed of overridden property Symfony\Component\Validator\Constraint::$errorNames.
```


enumag:
I don't think the failures are related to my change. :thinking:

ondrejmirtes:
Thank you.

PR #205, opened 2021-11-03
Fix: Incorrect return type for InputBag::all()

### LINKED ISSUE #194: Incorrect return type for InputBag::all()
#144 changed the return type of `InputBag::all()` to `string[]`. Unfortunately that is incorrect: It is, for example, possible to do something like this: `GET /?items[][id]=1` which will cause `InputBag::all('items')` to return `[0 => ['id' => '1']]`.

blankse:
Fix #194

### DIFF
--- a/src/Type/Symfony/InputBagDynamicReturnTypeExtension.php
+++ b/src/Type/Symfony/InputBagDynamicReturnTypeExtension.php
@@ -70,18 +70,11 @@ final class InputBagDynamicReturnTypeExtension implements DynamicMethodReturnTyp
 		MethodCall $methodCall
 	): Type
 	{
-		$types = [
-			new BooleanType(),
-			new FloatType(),
-			new IntegerType(),
-			new StringType(),
-		];
-		$oneParameterType = new UnionType($types);
 		if (isset($methodCall->getArgs()[0])) {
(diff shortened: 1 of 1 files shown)


src/Type/Symfony/InputBagDynamicReturnTypeExtension.php
```diff
@@ -70,18 +70,11 @@ private function getAllTypeFromMethodCall(
...
-		$types = [
-			new BooleanType(),
-			new FloatType(),
-			new IntegerType(),
-			new StringType(),
-		];
-		$oneParameterType = new UnionType($types);
 		if (isset($methodCall->getArgs()[0])) {
-			return new ArrayType(new MixedType(), $oneParameterType);
+			return new ArrayType(new MixedType(), new MixedType());
```
ondrejmirtes, line 74: Can you use `new MixedType(true)` so that it causes issues at least on level 9?

ondrejmirtes (changes requested):

src/Type/Symfony/InputBagDynamicReturnTypeExtension.php
blankse: @ondrejmirtes Changed it

PR #209, opened 2021-11-24
XmlServiceMapFactory: Add more info in case of XML can not be parsed.

janbarasek:
The current error message only informs that a parsing error has occurred. More information should be added to the message as to why the parsing failed.

Current error message:

Snímek obrazovky 2021-11-24 v 14 17 31


ondrejmirtes:
Hi, a test would be nice! :) It should be easy, a new `XmlServiceMapFactoryTest` that tries to call `create()` with a malformed file.

lookyman:
Also please add the same logic to `XmlParameterMapFactory`.

PR #210, opened 2021-11-29
Fix false positive "Service ... is private." when using KernelTestCase::$container

jeroennoten:
This fixes the false positive reported in #27.

This makes sure that when you call `self::$container->get(...)` in a `KernelTestCase`, you don't get the 'Service ... is private.' error.


ondrejmirtes:
Thank you!

PR #211, opened 2021-11-29
Services are private by default

### LINKED ISSUE #75: ContainerInterfacePrivateServiceRule should report services that are implicitly private
[Since Symfony 3.4, services are private by default](https://symfony.com/blog/new-in-symfony-3-4-services-are-private-by-default). However, the [`XmlServiceMapFactory`](https://github.com/phpstan/phpstan-symfony/blob/master/src/Symfony/XmlServiceMapFactory.php) still assumes that a service is public if the `public` attribute hasn't been set:

https://github.com/phpstan/phpstan-symfony/blob/a6d87e0068741327f99cf3ec0e9c4f6809e602e2/src/Symfony/XmlServiceMapFactory.php#L48-L54

([the 3rd argument of the `Service` constructor is `bool $public`](https://github.com/phpstan/phpstan-symfony/blob/a6d87e0068741327f99cf3ec0e9c4f6809e602e2/src/Symfony/Service.php#L26)).

This means that the [`ContainerInterfacePrivateServiceRule`](https://github.com/phpstan/phpstan-symfony/blob/master/src/Rules/Symfony/ContainerInterfacePrivateServiceRule.php) will only report private services if they are explicitly marked as private in their service configuration, but not if they are implicitly private.

In order to correctly report private services being fetched from the container, this logic should be reversed so that a service without a `public` flag is considered private:

```php
$service = new Service(
	// ...
	isset($attrs->public) && (string) $attrs->public !== 'false',
	// ...
);
```

That will result in false positives for Symfony versions < 3.4, [but those versions are no longer supported by Symfony itself](https://symfony.com/releases). Alternatively, the `XmlServiceMapFactory` would have to determine the Symfony version and switch it's logic based on that, but I'm not sure if that's even possible...

jeroennoten:
This fixes #75.

This changes marks services as private when they are not explicitly public.

### DIFF
--- a/src/Symfony/XmlServiceMapFactory.php
+++ b/src/Symfony/XmlServiceMapFactory.php
@@ -48,7 +48,7 @@ final class XmlServiceMapFactory implements ServiceMapFactory
 			$service = new Service(
 				strpos((string) $attrs->id, '.') === 0 ? substr((string) $attrs->id, 1) : (string) $attrs->id,
 				isset($attrs->class) ? (string) $attrs->class : null,
-				!isset($attrs->public) || (string) $attrs->public !== 'false',
+				isset($attrs->public) && (string) $attrs->public === 'true',
 				isset($attrs->synthetic) && (string) $attrs->synthetic === 'true',
 				isset($attrs->alias) ? (string) $attrs->alias : null
 			);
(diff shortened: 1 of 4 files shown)


ondrejmirtes:
Thank you!

Pixelshaped:
This needs to be reverted/fixed, it doesn't cover legitimate use cases that the official documentation covers, see #219

PR #212, opened 2021-11-29
Followup of Symfony TestContainer issue

Pierstoval:
Followup of #27 and #210

### DIFF
--- a/stubs/Symfony/Bundle/FrameworkBundle/Test/KernelTestCase.stub
+++ b/stubs/Symfony/Bundle/FrameworkBundle/Test/KernelTestCase.stub
@@ -8,4 +8,9 @@ abstract class KernelTestCase
      * @var TestContainer
      */
     protected static $container;
+
+    /**
+     * @return TestContainer
+     */
+    abstract public function getContainer();
 }


ondrejmirtes:
Please fix the build failures.

Pierstoval:
Should be done!

yakobe:
@Pierstoval This works perfectly for me too 👍 ! 
@ondrejmirtes Would it be possible to get this in? Or is there a release schedule to wait for?
Thanks for all the hard work everyone 😊.

ondrejmirtes:
Thank you! Gonna release it immediately.

alexander-schranz:
This seems to end in a strange error now when using a trait expecting a `getContainer` method with `ContainerInterface` as return type to exist:

I'm now getting:

```bash
 ------ ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
  Line   src/Sulu/Bundle/TestBundle/Testing/ContainerTrait.php (in context of class Sulu\Bundle\TestBundle\Testing\KernelTestCase)
 ------ ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
  32     Method Sulu\Bundle\TestBundle\Testing\KernelTestCase::getContainer() should return Symfony\Bundle\FrameworkBundle\Test\TestContainer but returns Symfony\Component\DependencyInjection\ContainerInterface.
 ------ ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
```

I'm not understanding why here `Symfony\Bundle\FrameworkBundle\Test\TestContainer` is expected.

I think the Problem is that phpstan is not knowing that `TestContainer` is an instance of `Symfony\Component\DependencyInjection\ContainerInterface` now? But I could not find it how to achieve this.

stubs/Symfony/Bundle/FrameworkBundle/Test/KernelTestCase.stub
```diff
@@ -8,4 +8,9 @@ abstract class KernelTestCase
      * @var TestContainer
      */
     protected static $container;
+
+    /**
+     * @return TestContainer
+     */
+    abstract public function getContainer();
```
alexander-schranz, line 15: Also think this should be `protected static` like the implemntation: https://github.com/symfony/symfony/blob/94533d419fb54aac8c9391fa9cbe03342ee8f374/src/Symfony/Bundle/FrameworkBundle/Test/KernelTestCase.php#L88

janedbal:
I'm meeting the same issue as @alexander-schranz, why `KernelTestCase::getContainer` is marked to return TestContainer? There is no such thing guarranteed in the [implementation](https://github.com/symfony/symfony/blob/94533d419fb54aac8c9391fa9cbe03342ee8f374/src/Symfony/Bundle/FrameworkBundle/Test/KernelTestCase.php#L88), is it?

Pierstoval:
> I'm meeting the same issue as @alexander-schranz, why `KernelTestCase::getContainer` is marked to return TestContainer? There is no such thing guarranteed in the [implementation](https://github.com/symfony/symfony/blob/94533d419fb54aac8c9391fa9cbe03342ee8f374/src/Symfony/Bundle/FrameworkBundle/Test/KernelTestCase.php#L88), is it?

Indeed there's no real guarantee based on types, but in practice, it's almost always the case when using the full-stack framework: https://github.com/symfony/symfony/blob/60ce5a3dfbd90fad60cd39fcb3d7bf7888a48659/src/Symfony/Bundle/FrameworkBundle/Resources/config/test.php#L48

alexander-schranz:
Think `Symfony\Bundle\FrameworkBundle\Test\TestContainer` is correct and nobody would overwrite that. But think the main problem is that PHPStan does not now that `Symfony\Bundle\FrameworkBundle\Test\TestContainer` is implementing `Symfony\Component\DependencyInjection\ContainerInterface`. Because if it would I think the error would be gone.

The `stubs` implementation looks for me currently not allowing to achieve this, and also think that the `stubs` are not very future proof as when typehints change in different symfony versions.  Maybe there is another workaround to tell that `KernelTestCase` return a `TestContainer` without adding stubs here, so PHPStan also knows which interfaces TestContainer implements and also would implement in the future. /cc @ondrejmirtes do you see other ways to achieve the same result?

Maybe a more general solution to provide access to test container xml and dev container xml and based what files are analyzed the correct services are returned so also other services which are available only in test environment work like expected:

```xml
        container_xml_path: var/cache/dev/srcDevDebugProjectContainer.xml
        test_container_xml_path: var/cache/test/srcTestDebugProjectContainer.xml
        test_file_regex: 'Test.php$'
```

Or we can manipulate the loading of `srcDevDebugProjectContainer` that `test.service_container` points to TestContainer.

Pierstoval:
`test.service_container` should already contain an instance of `TestContainer`.

About the XML part, I'm not sure about what you'd like to achieve. I personally specify the Test container in my `container_xml_path` and it doesn't detect the `TestContainer` either, unless using the stubs in phpstan 😕

PR #220, opened 2021-12-15
Fix false positive about private service in service subscriber

### LINKED ISSUE #219: Services are not private if included through a Service Subscriber/Locator
Since #211 there is no way to get Services on the container by default. This a nice idea but it overlooks legitimate use-cases: it is indeed possible to get services on the container without getting a deprecation warning if your service implements the `ServiceSubscriberInterface` for example (see docs: https://symfony.com/doc/current/service_container/service_subscribers_locators.html#defining-a-service-subscriber)

It's the preferred way to make loading of services lazy: "Sometimes, a service needs access to several other services without being sure that all of them will actually be used. In those cases, you may want the instantiation of the services to be lazy". If you want an official example, go look at [AbstractController](https://github.com/symfony/symfony/blob/807672d639a888b7c538a2c3b2a4632339f417cb/src/Symfony/Bundle/FrameworkBundle/Controller/AbstractController.php#L57)

Now this raises errors:
```
 ------ --------------------------------------------------------------------- 
  Line   src/Twig/CategoryExtension.php                                       
 ------ --------------------------------------------------------------------- 
  33     Service "App\Service\ProductCategoryHelper" is private.              
  33     Service "Symfony\Component\HttpFoundation\RequestStack" is private.  
 ------ --------------------------------------------------------------------- 
```

```
class CategoryExtension extends AbstractExtension implements ServiceSubscriberInterface
{
    public function __construct(
        private ContainerInterface $container
    ) {}

    public function getFunctions(): array
    {
        return [
            ... whole bunch of other functions ...
            new TwigFunction('get_category_tree', [$this, 'getCategoryTree']),
        ];
    }

    /**
     * @return Collection<int, Category>
     */
    public function getCategoryTree(): Collection
    {
        return $this->container->get(ProductCategoryHelper::class)->getCategoryTree($this->container->get(RequestStack::class)->getCurrentRequest()->getLocale());
    }

    /**
     * @return array<string>
     */
    public static function getSubscribedServices(): array
    {
        return [
            ProductCategoryHelper::class,
            RequestStack::class
        ];
    }
}
```
But really, it shouldn't.

@jeroennoten 

Current Symfony version: 5.4.1

jeroennoten:
This fixes #219

### DIFF
--- a/src/Rules/Symfony/ContainerInterfacePrivateServiceRule.php
+++ b/src/Rules/Symfony/ContainerInterfacePrivateServiceRule.php
@@ -8,7 +8,9 @@ use PHPStan\Analyser\Scope;
 use PHPStan\Rules\Rule;
 use PHPStan\ShouldNotHappenException;
 use PHPStan\Symfony\ServiceMap;
+use PHPStan\TrinaryLogic;
 use PHPStan\Type\ObjectType;
+use PHPStan\Type\Type;
 use function sprintf;
 
 /**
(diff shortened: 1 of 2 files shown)


ondrejmirtes:
Thank you!

PR #224, opened 2021-12-28
Replace tabs with 4 spaces in docs code blocks

bocharsky-bw:


lookyman:
What is the reason for this? The NEON format traditionally uses tabs.

ondrejmirtes:
Yeah, both work, there's no need to change this.

PR #225, opened 2021-12-30
Fix FormView stub

VincentLanglet:
Hi @ondrejmirtes, since https://github.com/symfony/symfony/pull/44764/files the FormView phpstan-stub is wrong.

This create an issue because when using both phpstan and psalm, the first one require string as key when the second use the symfony definition and allow both int and string.


lookyman:
LGTM 👍

ondrejmirtes:
Thank you!

PR #233, opened 2022-01-11
Request > Return `Session` instead of `SessionInterface`

ruudk:
Symfony 6 removed the `Session` and `FlashBagInterface` services.

That means that the only way to get the FlashBag (to add flashes) it to use `$request->getSession()->getFlashBag()`.

But the problem is, `getFlashBag` is not defined on the `SessionInterface`:
```
Call to an undefined method Symfony\Component\HttpFoundation\Session\SessionInterface::getFlashBag().
```

To make everyone's life easier, let's change the return type to `Sesssion` instead.
This is what's returned in a default Symfony application.

### DIFF
--- a/extension.neon
+++ b/extension.neon
@@ -118,6 +118,11 @@ services:
 		factory: PHPStan\Type\Symfony\RequestDynamicReturnTypeExtension
 		tags: [phpstan.broker.dynamicMethodReturnTypeExtension]
 
+	# Request::getSession() return type
+	-
+		factory: PHPStan\Type\Symfony\RequestGetSessionReturnTypeExtension
+		tags: [phpstan.broker.dynamicMethodReturnTypeExtension]
+
 	# Request::getSession() type specification
 	-
 		factory: PHPStan\Type\Symfony\RequestTypeSpecifyingExtension
--- a/src/Type/Symfony/RequestGetSessionReturnTypeExtension.php
+++ b/src/Type/Symfony/RequestGetSessionReturnTypeExtension.php
@@ -1,0 +1,34 @@
+<?php declare(strict_types = 1);
+
+namespace PHPStan\Type\Symfony;
+
+use PhpParser\Node\Expr\MethodCall;
+use PHPStan\Analyser\Scope;
+use PHPStan\Reflection\MethodReflection;
+use PHPStan\Type\DynamicMethodReturnTypeExtension;
+use PHPStan\Type\ObjectType;
+use PHPStan\Type\Type;
+
+final class RequestGetSessionReturnTypeExtension implements DynamicMethodReturnTypeExtension
+{
+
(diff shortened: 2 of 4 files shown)


ruudk:
@ondrejmirtes Does this make sense? Or should it be done differently?

lookyman:
It makes sense, but still, it's technically straight up lying about the return type. I don't really like it.

ruudk:
But it means that all code needs to do an assertion check. For example, right now I have 40 errors because getFlashBag does not exist on session on request. 

I thought these extensions should solve friction like this.

lookyman:
Can we somehow find out about the real return type from the service map?

ruudk:
More info on this matter: 
- https://github.com/symfony/symfony/issues/41765
- https://github.com/symfony/symfony/issues/39222

ruudk:
I think the suggested change makes a lot of sense because:
* this is a symfony plugin; 
* almost all symfony users have the normal session configuration, meaning request session returns the session object.

For a small set of users it _could_ mean that phpstan now falsely allows getFlashBag on the request session, because they have a custom session implementation. They prob didn't call the flashbag anyway, as it doesn't work. And if they implemented it, great. If needed, and those complain, we can add a toggle to disable it for them.

lookyman:
You know what, after some digging in Symfony's code, I think you are right. Lets merge this.

ondrejmirtes:
Hi, I reused your test. Since the return type wasn't really dynamic, the same result can be achieved with stub files: https://github.com/phpstan/phpstan-symfony/commit/012305dab7a08f4e1a7158e01d7b4bccfbc6fa31

Thank you.

ruudk:
Interesting, that was my first try but it didn't seem to pick it up. I read the code as : if request class does not exist then use stub

ondrejmirtes:
Nope, that's not how stub files work. They're purely for overriding PHPDocs, nothing else.

VincentLanglet:
@ondrejmirtes Hi, I'm curious ; is it possible to make this return type a `Benevolent<Session|SessionInterface>` with phpdoc ?

ondrejmirtes:
If Session implements SessionInterface, then `Session|SessionInterface` is just `SessionInterface`. Thanks to https://phpstan.org/developing-extensions/type-system#type-normalization.

VincentLanglet:
> If Session implements SessionInterface, then `Session|SessionInterface` is just `SessionInterface`. Thanks to https://phpstan.org/developing-extensions/type-system#type-normalization.

Oh I see...

So there is no way to add phpdoc in a way that
- `$session->getFlashbag()` is not reported because it exists in Session
- `$session instanceof Session` is not reported as always true because it can be a SessionInterface.

ondrejmirtes:
Both can't be true at the same time.

PR #234, opened 2022-01-16
Post request variables can be only string

ossinkine:
Just like query variables the request variables can be only string. Even you sends `1` as post variable, you gets string.

@ondrejmirtes Why was it originally done differently?


ondrejmirtes:
This commit in Symfony specifically added int, float, and bool because of POST requests, so there's probably a way: https://github.com/symfony/symfony/commit/381a0a19f741da314dff99fc64cc75d674fc7beb

Issue #235, opened 2022-01-17
CacheInterface->get returns mixed


YspMark:
The `CacheInterface` `get` method is by default hinted with the `mixed` return type. The `callback` parameter return will be the cache item, so that return type could be used to determine the right return type.

PR #236, opened 2022-01-17
Added CacheInterface stub

YspMark:
Closes https://github.com/phpstan/phpstan-symfony/issues/235


lookyman:
The `callable` annotation has to have parentheses for parameters, event if there are none. Like this `callable(): T`.

However, in this case the callback actually accepts `Psr\Cache\CacheItemInterface` as the first (and only) parameter.

Also, the parameter is annotated as `callable|CallbackInterface` in Symfony. So I think we need to add a second stub for the `CallbackInterface`, with templated `__invoke` method, and the one for `CacheInterface`'s `$callback` needs to be annotated something like this: `@param \Symfony\Contracts\Cache\CallbackInterface<T>|callable(\Psr\Cache\CacheItemInterface): T $callback`.

stubs/Symfony/Contracts/Cache/CacheInterface.stub
```diff
@@ -0,0 +1,15 @@
+<?php
+
+namespace Symfony\Contracts\Cache;
+
+interface CacheInterface
+{
+    /**
+     * @template T
+     *
+     * @param \Symfony\Contracts\Cache\CallbackInterface<T>|callable(\Psr\Cache\CacheItemInterface): T $callback
```
lookyman, line 10: `\Psr\Cache\CacheItemInterface=`, notice the `=` at the end. See https://phpstan.org/writing-php-code/phpdoc-types#callables.

For the `Parameter $callback of method Symfony\Contracts\Cache\CacheInterface::get() has invalid type Psr\Cache\CacheItemInterface.` error, we need to actually add the `Psr\Cache\CacheItemInterface` stub as well. It should be sufficient for it to be an empty class. We only need to know the class, we don't care about methods. See https://phpstan.org/user-guide/stub-files, `Why is PHPStan complaining about “Class not found” in a stub file?`

YspMark: Thanks for the feedback, I was not familiar with the `= optional` notation, actually thought it was a typo.

ondrejmirtes:
Can you please write some tests in https://github.com/phpstan/phpstan-symfony/blob/master/tests/Type/Symfony/ExtensionTest.php to verify this signature works as expected? More info here: https://phpstan.org/developing-extensions/testing#type-inference

ondrejmirtes (changes requested):

ondrejmirtes (approved):

ondrejmirtes:
Thank you.

PR #238, opened 2022-01-20
Add analysis for route generation by controllers and UrlGeneratorInterface

DaDeather:
Hi there.

we were in need of this feature to detect possible dead route generation when using dynamicly generated route "name"s through symfony by its `namespace_controller_method` style.

Therefore I've created this simple analysis which I would like to share with anyone in the same need.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,6 +23,7 @@ This extension provides following features:
 * Provides correct return types for `TreeBuilder` and `NodeDefinition` objects.
 * Notifies you when you try to get an unregistered service from the container.
 * Notifies you when you try to get a private service from the container.
+* Notifies you when you try to generate a URL for a non-existing route name.
 * Optionally correct return types for `InputInterface::getArgument()`, `::getOption`, `::hasArgument`, and `::hasOption`.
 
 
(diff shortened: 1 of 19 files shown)


DaDeather:
@ondrejmirtes would this feature be something you would see in this project?

ondrejmirtes:
I need @lookyman to look at this, I'm not able to evaluate it myself as I personally don't use Symfony. Thanks.

DaDeather:
@lookyman did you had any chance to look over it or at least for letting me know if this might be a feature of relevance for this plugin?

lookyman:
I did have a look. It's an OK feature, but to be honest, it's pretty specific use case, and I'm not sure I really want to maintain it here..

DaDeather:
> I did have a look. It's an OK feature, but to be honest, it's pretty specific use case, and I'm not sure I really want to maintain it here..

It is definitely a specific feature and I fully understand the maintaining issue here. I will outsource it to another repository than to allow others to use / copy / extend it by need.
Thank you guys for all your work!

DaDeather:
For anyone interested in such a feature - here's the package:
https://packagist.org/packages/dadadev/phpstan-symfony-routing

PR #240, opened 2022-01-24
Add missing save parameter to CacheInterface stub

Khartir:
The stub is missing the second save parameter on the callable that is defined in the [CallbackInterface](https://github.com/phpstan/phpstan-symfony/blob/96ee630098c49c1c25326ae51b027e469bfd2064/stubs/Symfony/Contracts/Cache/CallbackInterface.stub)


ondrejmirtes:
Thank you!

PR #250, opened 2022-02-02
Remove nullable type on `InputBag::get()` after `InputBag::has()` call

franmomu:
I've used https://github.com/phpstan/phpstan-symfony/blob/df5ce326a8935bf400eb59f7080e568c20ef6246/src/Type/Symfony/RequestTypeSpecifyingExtension.php as example.

**Update**: maybe we could change it to a `HasGetTypeSpecifyingExtension` and pass the `className` so it can be reused for other classes as well.

### DIFF
--- a/extension.neon
+++ b/extension.neon
@@ -278,3 +278,8 @@ services:
 	-
 		factory: PHPStan\Type\Symfony\ResponseHeaderBagDynamicReturnTypeExtension
 		tags: [phpstan.broker.dynamicMethodReturnTypeExtension]
+
+	# InputBag::get() type specification
+	-
+		factory: PHPStan\Type\Symfony\InputBagTypeSpecifyingExtension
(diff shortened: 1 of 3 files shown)


lookyman:
👍

ondrejmirtes:
Thank you!

VincentLanglet:
Hi @franmomu, I'm not sure this DynamicReturnExtension is correct, 

The `has` method is using the `array_key_exists` check
https://github.com/symfony/symfony/blob/6.3/src/Symfony/Component/HttpFoundation/ParameterBag.php#L99

Which means that we can have
```
$bag->set('foo', null);
$bag->has('foo'); // true
$bag->get('foo'); // null
```

If so, I'm affraid we'll have to remove this DynamicReturnExtension :(
WDYT ?

PR #254, opened 2022-02-09
Support for parameterised service class

### LINKED ISSUE #226: Services with parameterised `class` are not resolved
I've installed `phpstan/phpstan-symfony` extension and analysed current codebase with configured container. I see many errors like `Property Foo::$bar (Bar) does not accept %foo.bar.service.class%` on `$this->bar = $this->container->get('foo.bar.service');` when service has definition with parameterised class:

```
foo.bar.service:
  class: '%foo.bar.service.class%'
```

Such service definition is totally correct. It seems like it can't be resolved properly.

FYI: We don't have full Symfony framework, we use `symfony/dependency-injection` and dump container similarly like framework does. I don't know if that matters, but I thought I should point it out.

Wirone:
Fixes #226

~~It's simplest and optimistic implementation which does not support:~~
- ~~`%env()%` syntax~~
- ~~`%app.class%_decorator` mixed interpolation~~

Let me know if we can start with this or should we work on most advanced scenarios 🙂

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,6 +18,7 @@
     "phpstan/phpstan": "^1.4"
   },
   "conflict": {
+    "psr/container": "1.1.2",
     "symfony/framework-bundle": "<3.0"
   },
   "require-dev": {
@@ -28,6 +29,7 @@
     "phpunit/phpunit": "^9.5",
     "symfony/config": "^4.2 || ^5.0",
     "symfony/console": "^4.0 || ^5.0",
+    "symfony/dependency-injection": "^4.0 || ^5.0",
     "symfony/form": "^4.0 || ^5.0",
     "symfony/framework-bundle": "^4.4 || ^5.0",
     "symfony/http-foundation": "^5.1",
--- a/composer.json
+++ b/composer.json
@@ -26,7 +26,7 @@
     "phpstan/phpstan-phpunit": "^1.0",
     "phpstan/phpstan-strict-rules": "^1.0",
     "phpunit/phpunit": "^9.5",
-    "psr/container": "^1.1 <1.1.2",
+    "psr/container": "1.0 || 1.1.1",
     "symfony/config": "^4.2 || ^5.0",
     "symfony/console": "^4.0 || ^5.0",
     "symfony/dependency-injection": "^4.0 || ^5.0",
--- a/composer.json
+++ b/composer.json
@@ -18,7 +18,6 @@
     "phpstan/phpstan": "^1.4"
   },
   "conflict": {
-    "psr/container": "1.1.2",
     "symfony/framework-bundle": "<3.0"
   },
   "require-dev": {
@@ -27,6 +26,7 @@
     "phpstan/phpstan-phpunit": "^1.0",
     "phpstan/phpstan-strict-rules": "^1.0",
     "phpunit/phpunit": "^9.5",
+    "psr/container": "^1.1 <1.1.2",
     "symfony/config": "^4.2 || ^5.0",
     "symfony/console": "^4.0 || ^5.0",
     "symfony/dependency-injection": "^4.0 || ^5.0",
--- a/src/Type/Symfony/ServiceDynamicReturnTypeExtension.php
+++ b/src/Type/Symfony/ServiceDynamicReturnTypeExtension.php
@@ -8,12 +8,17 @@ use PHPStan\Reflection\MethodReflection;
 use PHPStan\Reflection\ParametersAcceptorSelector;
 use PHPStan\ShouldNotHappenException;
 use PHPStan\Symfony\Configuration;
+use PHPStan\Symfony\ParameterMap;
+use PHPStan\Symfony\ServiceDefinition;
 use PHPStan\Symfony\ServiceMap;
 use PHPStan\Type\Constant\ConstantBooleanType;
 use PHPStan\Type\DynamicMethodReturnTypeExtension;
 use PHPStan\Type\ObjectType;
 use PHPStan\Type\Type;
 use function in_array;
+use function is_scalar;
+use function strpos;
+use function trim;
 
 final class ServiceDynamicReturnTypeExtension implements DynamicMethodReturnTypeExtension
 {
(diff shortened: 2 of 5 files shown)


ondrejmirtes:
/cc @lookyman needs to say if it's okay :)

This is how the `%` expanding logic looks like in nette/di: https://github.com/nette/di/blob/5ef440ce8c603bb80a4a6459e0842528c5bf64ea/src/DI/Helpers.php#L26-L108

Can't we just call something in Symfony to achieve the same?

Wirone:
Yeah, it would be great to use some kind of resolver and I'll look at it if needed, I just wanted to know if I hooked into right place with parameter expanding 🙂 So let me know @lookyman if that's correct, so I can work on `ServiceDynamicReturnTypeExtension::determineServiceClass()` then.

lookyman:
Seems ok to me, go ahead.

Wirone:
@lookyman Symfony resolves placeholders in `Symfony\Component\DependencyInjection\Compiler\ResolveParameterPlaceHoldersPass` when DI container is compiled. Since this extension does not compile container but uses only simple maps (separate for services and parameters) I decided to use `ParameterBag` which has method for resolving values. So if service's definition uses placeholders, these will be resolved into actual values using parameter bag.

I did not want to create another `ParameterMap` implementation, but if you don't like current implementation just let me know how would you want it to be.

PS. `make` locally passes, so Github actions' fail is not related to changes.

Wirone:
@ondrejmirtes @lookyman In order to support PHP7.1 I had to enforce `symfony/dependency-injection` v4, which caused also `symfony/framework-bundle` to downgrade. Unfortunately there are still some dependencies not resolved properly, causing problems on 7.1. Could you help me with providing proper Composer requirements? Maybe you had similar problems before 🙂

Wirone:
Sorry for the noise, but it took me while to understand why it fails on PHP 7.1. Personally I think it's wrong that packages used in actual code are added as dev dependencies, but now I know that adding them to `require` causes trouble 😉 These packages now are soft dependencies but probably you assume that they're required by project that uses PHPStan with this extension and will be installed anyway, so analysis will work. Anyway, checks are green, parameter resolving works as expected - my job is done here 🙂

composer.json
```diff
@@ -18,6 +18,7 @@
     "phpstan/phpstan": "^1.4"
   },
   "conflict": {
+    "psr/container": "1.1.2",
```
ondrejmirtes, line 21: Why the conflict?

ondrejmirtes: Can we bump version of something in Symfony in require-dev instead of conflicting with this?

Wirone: It does not relate to Symfony but to [PHP versions](https://github.com/php-fig/container/issues/30#issuecomment-791556780). After importing `Symfony\Component\DependencyInjection\ParameterBag\ParameterBag` in `ServiceDynamicReturnTypeExtension` it implicitly imported Symfony DI exceptions and problem appeared.

The bug was already here before my changes, but code was never used and analysed, hence green checks, but if you open `Symfony\Component\DependencyInjection\Exception\ExceptionInterface` in IDE on `master` branch you'll see:

Symfony\Component\DependencyInjection\Exception\ExceptionInterface cannot implement previously implemented interface Throwable

ondrejmirtes: You can put newer "psr/container" in require-dev instead.

Wirone: @ondrejmirtes can you tell me what would be the difference? 🤔  `psr/container` is required anyway:

```
$ composer why psr/container
symfony/dependency-injection  v5.4.3  requires  psr/container (^1.1.1)
symfony/service-contracts     v2.5.0  requires  psr/container (^1.1)
```

adding it to `require-dev` won't change anything since I have to add constraint `^1.1 <1.1.2` and restrict installation of 1.1.1 anyway. `conflicts` does the same but on different level of package resolution... or am I missing something?

ondrejmirtes: The difference is that I don't want the conflict for the consumers of this package. I want them to be able to use it because the problematic part might not even be executed in their use-case.

Wirone: OK, I'll change it.

-----

But IMHO using `require-dev` here is only dirty hack. In fact, e.g. `symfony/dependency-injection` is used in the actual code (not in dev processes only) and it should be strict requirement, installable on consumers' side even without explicit requirement in their `composer.json`. Right now you only assume that those packages will be installed. [Composer Require Checker](https://github.com/maglnet/ComposerRequireChecker) reports that 56 unknown symbols were found.

Also, you support PHP7.1 in Github Actions' matrix while there's (dev) requirement for `symfony/http-foundation ^5.1` which requires PHP >=7.2.5... Looking at the [changelog](https://github.com/symfony/http-foundation/blob/5.4/CHANGELOG.md#500) and usages of `Symfony\Component\HttpFoundation` namespace in this package there's probably no problem if consumer has `symfony/http-foundation` v4, but in general it's not good that Symfony extension uses Symfony components without requiring them explicitly. In fact, this works in Github checks only because of [platform overriding](https://github.com/phpstan/phpstan-symfony/blob/master/composer.json#L40), but on actual PHP7.1 it's not installable:

```
/app # composer require --dev phpunit/phpunit:^7.5.20 -w
./composer.json has been updated
Running composer update phpunit/phpunit --with-dependencies
Loading composer repositories with package information
Updating dependencies
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - symfony/http-foundation[v5.1.0-BETA1, ..., v5.1.0-RC1] require php ^7.2.5 -> your php version (7.1.33) does not satisfy that requirement.
    - symfony/http-foundation[v5.1.0-RC2, ..., 5.4.x-dev] require php >=7.2.5 -> your php version (7.1.33) does not satisfy that requirement.
    - Root composer.json requires symfony/http-foundation ^5.1 -> satisfiable by symfony/http-foundation[v5.1.0-BETA1, ..., 5.4.x-dev].
```

Also usage of Symfony packages v6 is not restricted on consumer's side (because you require `^4.2 || ^5.0` only for development) and I'm not sure if it'll work properly 🤔 

Sorry for this rant 😅 I don't want to blame you for anything, I am just wondering if this is proper approach in terms of QA for this package. Feel free to ignore it in terms of this MR, but I really would like to know your opinion and reasoning behind current architecture.

Wirone:
@ondrejmirtes @lookyman any chance for merging this?

ondrejmirtes:
Thank you.

Seldaek:
This created a regression if you're not using a full symfony framework. I use this project in Composer for example and I don't have the DI component installed as I only need the symfony/console type inferrence.


See https://github.com/composer/composer/runs/5334439438?check_suite_focus=true#step:9:10

ondrejmirtes:
@Seldaek Sorry for the bug, here's a fix: https://github.com/phpstan/phpstan-symfony/commit/8ef80f5c4f70ebd31a55647bc082ea9df4bf0132

Please test phpstan-symfony 1.1.7 with it, thank you. https://github.com/phpstan/phpstan-symfony/releases/tag/1.1.7

Seldaek:
It builds again, thanks for the super quick fix! https://github.com/composer/composer/actions/runs/1900486331

Wirone:
Sorry @Seldaek for causing problems 😳 And thanks @ondrejmirtes for the fix 🍻

Seldaek:
No worries, it's easy to forget these components are used without the full framework sometimes :)

PR #259, opened 2022-03-02
create FormErrorIterator stub

rgrassian:
Hi @ondrejmirtes !

see: https://github.com/symfony/form/commit/18f4f8209d2db06fc3cf5ba3b2eb8e4706b34aac

### DIFF
--- a/stubs/Symfony/Component/Form/FormErrorIterator.stub
+++ b/stubs/Symfony/Component/Form/FormErrorIterator.stub
@@ -1,0 +1,27 @@
+<?php
+
+namespace Symfony\Component\Form;
+
+/**
+ * @template T of FormError|FormErrorIterator
+ * @phpstan-implements \ArrayAccess<int, T>
+ * @phpstan-implements \RecursiveIterator<int, T>
+ * @phpstan-implements \SeekableIterator<int, T>
+ */
+class FormErrorIterator implements \RecursiveIterator, \SeekableIterator, \ArrayAccess, \Countable
+{
+    /**
+     * @param array<T> $errors
+     */
+    public function __construct(FormInterface $form, array $errors);
+
+    /** @return T */
+    public function current();
+
+    /**
+     * @param mixed $position
+     *
+     * @return T
+     */
+    public function offsetGet($position);
+}
--- a/extension.neon
+++ b/extension.neon
@@ -26,6 +26,8 @@ parameters:
 		- stubs/Symfony/Component/EventDispatcher/GenericEvent.stub
 		- stubs/Symfony/Component/Form/DataTransformerInterface.stub
 		- stubs/Symfony/Component/Form/FormBuilderInterface.stub
+		- stubs/Symfony/Component/Form/FormError.stub
+		- stubs/Symfony/Component/Form/FormErrorIterator.stub
 		- stubs/Symfony/Component/Form/FormInterface.stub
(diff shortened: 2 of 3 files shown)


VincentLanglet (approved):
We consider FormErrorIterator as a Generic in order to use the feature introduce in https://github.com/phpstan/phpstan-symfony/commit/29ad8a949139ba8a5342cda60c62fea4d7afafce
but the stub to declare this class as Generic was missing.

Since now Symfony added some phpdoc, there is some weird error like `getErrors` return FormErrorIterator<FormError> but method expect iterable<FormError>&FormErrorIterator.

In psalm, they added a stub for the FormErrorIterator ; this seems good to me to do the same here with this PR.

VincentLanglet:
I don't understand why the build is failing @ondrejmirtes, the `phpstan-implements` seems correct to me...

ondrejmirtes:
The last error output is irrelevant - PHPStan is getting killed because of some infinite loop/infinite recursion problem.

stubs/Symfony/Component/Form/FormErrorIterator.stub
```diff
@@ -0,0 +1,27 @@
...
+    /**
+     * @param array<T> $errors
+     */
+    public function __construct(FormInterface $form, array $errors);
+
+    /** @return T */
+    public function current();
+
+    /**
+     * @param int $position
```
VincentLanglet, line 22:
```suggestion
     * @param mixed $position
```
I think phpstan will prefer this

VincentLanglet:
> The last error output is irrelevant - PHPStan is getting killed because of some infinite loop/infinite recursion problem.

What about those errors @ondrejmirtes 
```
Class Symfony\Component\Form\FormErrorIterator implements generic   
         interface ArrayAccess but does not specify its types: TKey, TValue  
Class Symfony\Component\Form\FormErrorIterator implements generic   
         interface RecursiveIterator but does not specify its types: TKey,   
         TValue                                                              
Class Symfony\Component\Form\FormErrorIterator implements generic   
         interface SeekableIterator but does not specify its types: TKey,    
         TValue 
```

There is the phpdoc
```
 * @phpstan-implements \ArrayAccess<int, T>
 * @phpstan-implements \RecursiveIterator<int, T>
 * @phpstan-implements \SeekableIterator<int, T>
```

ondrejmirtes:
Again - The last error output is irrelevant - PHPStan is getting killed because of some infinite loop/infinite recursion problem.

VincentLanglet:
> Again - The last error output is irrelevant - PHPStan is getting killed because of some infinite loop/infinite recursion problem.

Sorry, I thought you were only talking about `Child process error (exit code 139): Segmentation fault (core dumped)`

rgrassian:
Hi @ondrejmirtes, I don't really understand, should I change something ? or it will be merged like that ?

ondrejmirtes:
I can't merge this, because it makes PHPStan crash. Someone has to investigate why, and fix the root issue :)

VincentLanglet:
> I can't merge this, because it makes PHPStan crash. Someone has to investigate why, and fix the root issue :)

Seems related to `@template T of FormError|FormErrorIterator`.

If I change to `@template T of FormError` or `@template T of mixed`, it doesn't crash.

Maybe it's similar to https://github.com/phpstan/phpstan/issues/6501 (?)
cc @arnaud-lb

PR #263, opened 2022-03-22
Update DenormalizerInterface::denormalize

VincentLanglet:
Hi @ondrejmirtes

The stub from phpstan-symfony is restricting the return type to object|array, but there is no restriction in the interface from Symfony ; cf https://github.com/symfony/Serializer/blob/6.1/Normalizer/DenormalizerInterface.php.

So I changed the return type to `mixed` like the one from Symfony.


VincentLanglet:
For more context @lucassabreu, I have a friend returning other thing than object/array in this method, and everything working fine.
So since even the phpdoc from symfony is not restricting to object/array, I don't think we should here.

lucassabreu (approved):

PR #267, opened 2022-04-05
Add missing Request.stub to extension.neon

t-richard:
On our project, we were hoping for this feature be available https://github.com/phpstan/phpstan-symfony/commit/012305dab7a08f4e1a7158e01d7b4bccfbc6fa31

But it's not by default because the extension.neon file doesn't contain this stub by default.

This PR fixes it.


t-richard:
Apparently this stub should be loaded by `InputBagStubFilesExtension` but isn't in our case.

I'll dig out why but this is not a proper fix for it anyway.

I'll close this, sorry for the noise.

PR #271, opened 2022-04-29
add documentation to load DI helper

### LINKED ISSUE #269: PHPstan does not find container helper functions
After upgrading to PHPStan 1.6, the following errors are reported when analyzing the config files:
```
Used function Symfony\Component\DependencyInjection\Loader\Configurator\service not found.
```
Workaround: Add `vendor/symfony/dependency-injection/Loader/Configurator/ContainerConfigurator.php` to the `scanFiles`.

I'm unsure if this is a bug or just something that should be added to the README.

Khartir:
Fixes #269, https://github.com/phpstan/phpstan/issues/7118.

As mentioned in https://github.com/symfony/symfony/pull/46214 this will not be changed in Symfony, so I added it to the documentation here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,6 +69,9 @@ parameters:
     # If you're using PHP config files for Symfony 5.3+, you also need this for auto-loading of `Symfony\Config`:
     scanDirectories:
         - var/cache/dev/Symfony/Config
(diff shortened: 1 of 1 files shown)


ondrejmirtes:
Thank you.

PR #274, opened 2022-05-27
Fix formInterface stub

VincentLanglet:
Hi @ondrejmirtes,  

A formInterface is not `\Traversable<int|string, \Symfony\Component\Form\FormInterface>` but `\Traversable<string, \Symfony\Component\Form\FormInterface>`

From https://github.com/symfony/symfony/blob/6.1/src/Symfony/Component/Form/FormInterface.php#L21-L24

### DIFF
--- a/stubs/Symfony/Component/Form/FormInterface.stub
+++ b/stubs/Symfony/Component/Form/FormInterface.stub
@@ -3,9 +3,10 @@
 namespace Symfony\Component\Form;
 
 /**
- * @extends \Traversable<string|int, \Symfony\Component\Form\FormInterface>
+ * @extends \ArrayAccess<string, \Symfony\Component\Form\FormInterface>
+ * @extends \Traversable<string, \Symfony\Component\Form\FormInterface>
  */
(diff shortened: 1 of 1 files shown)


ondrejmirtes:
Thank you.

VincentLanglet:
@ondrejmirtes, I recently encounter an issue related to this stub change.
You can see https://github.com/symfony/symfony/issues/46698 for more information.

This is a lot related to the fact that array-key like `'1'` are automatically cast to int by php.
It seems like Symfony often use such name as key. 

An exemple, code like
```
foreach ($form as $name => $child) {
     $form->remove($name);
}
```
fail with strict types because the $name is sometimes an int and `remove` expects a string.

So I start to wonder if it's not better to keep
```
\Traversable<string|int, \Symfony\Component\Form\FormInterface>
```
in order to prevent people "You might need to cast keys to string".

WDYT if I revert this PR with a comment ?

VincentLanglet:
Will be solved in 6.2 by https://github.com/symfony/symfony/pull/46813

PR #275, opened 2022-05-28
Try complex container param

VincentLanglet:


VincentLanglet:
@ondrejmirtes Would it be possible to not generalize too much array container params when resolving the values ?

I'd prefer to have
```
array{name: Vincent, data: array{foo: bar}}
```
generalized to
```
array{name: string, data: array{foo: string}}
```
rather than
```
array<string, string|array<string, string>>
```

Are you ok with this change ?
Any advice about how generalizeTypeFromValue and generalizeType should be updated ?

ondrejmirtes:
Yeah, I think it's fine for "indexed" arrays (string keys), but not for "lists" (int keys). I guess you have to write the generalization logic yourself.

VincentLanglet:
> Yeah, I think it's fine for "indexed" arrays (string keys), but not for "lists" (int keys). I guess you have to write the generalization logic yourself.

Done ! That was easier than expected :)

src/Type/Symfony/ParameterDynamicReturnTypeExtension.php
```diff
@@ -139,16 +141,30 @@ private function getGetTypeFromMethodCall(
...
-				new ArrayType(
-					TypeCombinator::union(...array_map(function ($item) use ($scope): Type {
-						return $this->generalizeTypeFromValue($scope, $item);
-					}, array_keys($value))),
-					TypeCombinator::union(...array_map(function ($item) use ($scope): Type {
-						return $this->generalizeTypeFromValue($scope, $item);
-					}, array_values($value)))
-				)
-			);
+			if (array_is_list($value)) {
```
ondrejmirtes: I'd say not to use these functions, although it might work because PHPStan includes these polyfills as well.

One thing you need to handle here is yes/no/maybe:

1) yes - array is list, array should be generalized
2) no - array is indexed, do the logic as you described
3) keys are mixed, some are integers, some are strings - array should also be generalized

Please also add a test for the scenario 3).

ondrejmirtes (changes requested):

src/Type/Symfony/ParameterDynamicReturnTypeExtension.php
VincentLanglet: > I'd say not to use these functions, although it might work because PHPStan includes these polyfills as well.

So how do you prefer I test if it's a list ?

ondrejmirtes: Iterate over the keys, see if all are integers or if all are strings, or if its a mix of strings/integers.

VincentLanglet: Sure, it's done :)

ondrejmirtes (approved):

PR #276, opened 2022-05-28
Try to support union of string as param

VincentLanglet:


VincentLanglet:
@ondrejmirtes This allow to resolve things like
```
$this->getParameterBagService()->get('webpack_encore.builds.' . $package)
```
if the `$package` can be a list of constant string.

In the second commit I tried to use a BC-way. Please tell me if you want that I do things differently.

It might also be possible to do it for the serviceId resolver. I prefer to do it in another PR and this might be less useful.

src/Symfony/ParameterMap.php
```diff
@@ -15,6 +18,15 @@ public function getParameters(): array;
...
+	/**
+	 * @deprecated Will be removed in 2.0
+	 */
 	public static function getParameterKeyFromNode(Expr $node, Scope $scope): ?string;
 
+	//  TODO: Uncomment this in 2.0
+	//	/**
+	//	 * @return array<string>
+	//	 */
+	//	public static function getParameterKeysFromNode(Expr $node, Scope $scope): array;
```
ondrejmirtes: I wouldn' be concerned with BC here, I don't think anyone implements this interface in their own code...

ondrejmirtes (changes requested):

src/Symfony/ParameterMap.php
VincentLanglet: Great, I reverted the BC commit then :)

src/Type/Symfony/ParameterDynamicReturnTypeExtension.php
```diff
@@ -187,18 +194,31 @@ private function getHasTypeFromMethodCall(
...
+		$has = null;
+		foreach ($parameterKeys as $parameterKey) {
 			$parameter = $this->parameterMap->getParameter($parameterKey);
-			return new ConstantBooleanType($parameter !== null);
+
+			if ($has === null) {
+				$has = $parameter !== null;
+			} elseif (
+				$has === true && $parameter === null
+				|| $has === false && $parameter !== null
```
ondrejmirtes: Parentheses here please to disambiguate the logic.

ondrejmirtes (changes requested):

src/Type/Symfony/ParameterDynamicReturnTypeExtension.php
VincentLanglet: Sure. Done.

PR #280, opened 2022-06-03
ParameterDynamicReturnTypeExtension: Only generalizeType for keys

VincentLanglet:


VincentLanglet:
Hi @ondrejmirtes, I just tried https://github.com/phpstan/phpstan-symfony/pull/275 release in my project and it seems like it doesn't cover all my use cases.

Sometimes, I have config this way:
```
foo:
    2000: 
         key: 'foo'
         value: 'bar'
    2001:
         key: 'foo2'
         value: 'bar2'
```
Since numeric keys as casted to int, the previous PR doesn't work on this config and the array is generalize to
```
array<int, array<string, string>>
```

Also with
```
foo:
    default: 
         key: 'key'
         value: 'value'
    2000: 
         key: 'foo'
         value: 'bar'
    2001:
         key: 'foo2'
         value: 'bar2'
```
Since the key are int|string.

I updated the logic to support those config.
IMHO, we just have to generalize for list of values. (Indexed by 0, 1, 2, 3, ...).

ondrejmirtes:
No, I think that you need to apply the logic recursively. Your first example should be generalized to `array<int, array{key: string, value: string}>`.

The second example to `array<int|string, array{key: string, value: string}>`

VincentLanglet:
> No, I think that you need to apply the logic recursively.
> 
> The second example to `array<int|string, array{key: string, value: string}>`

What do you mean ? `generalizeTypeFromValue` is already applied recursively.

> Your first example should be generalized to array<int, array{key: string, value: string}>.
> The second example to array<int|string, array{key: string, value: string}>

I have more complex examples like:
```
foo:
    default: 'value'
    2000: 
         key: 'foo'
         value: 'bar'
    2001:
         key: 'foo2'
         value: 'bar2'
```
or
```
foo:
    default: 'value'
    2000: 
         key: 'foo'
         value: 'bar'
    2001:
         key: 'foo2'
         value: 'bar2'
```
And I'd like to avoid the foo config to be generalized to `array<string|int, string|array{key: string, value: string}>`

Or maybe a simpler one:
```
value_by_id:
    ab23de: 'toto'
    1ae2ee: 'foo'
     '1111111': 'bar'
```
Consider that all three values are random string validated by the regex `[a-e0-9]^6`. One of them is only made with int (by luck) but I still want to consider it as a string. And using `'111111'` doesn't change anything since the key is casted to an int by php.

When key are provided they should matter and be respected by phpstan. When key are not provided, this is a list and it seems ok to generalize to `int`.

ondrejmirtes:
> And I'd like to avoid the foo config to be generalized to array<string|int, string|array{key: string, value: string}>

What? That's exactly what I'd expect :) Messy config leads to messy results...

VincentLanglet:
> > And I'd like to avoid the foo config to be generalized to array<string|int, string|array{key: string, value: string}>
> 
> What? That's exactly what I'd expect :) Messy config leads to messy results...

I updated the PR.

ondrejmirtes:
Thank you.

PR #282, opened 2022-06-07
Fix HeaderBag::all return value

### LINKED ISSUE #264: Add support for HeaderBag::all()
Since Symfony 5.4 all() has a generic return type defined which is nice, but sadly due to the $key arg it is very variable what you get back: https://github.com/symfony/symfony/blob/f5a8916bad0e45ef125d128dca6c2793bd0c55d2/src/Symfony/Component/HttpFoundation/HeaderBag.php#L66-L68

It'd be great if this extension supported all() like it does get(), switching to the correct return type based on $key being present or not.

Seldaek:
fixes #264

### DIFF
--- a/stubs/Symfony/Component/HttpFoundation/HeaderBag.stub
+++ b/stubs/Symfony/Component/HttpFoundation/HeaderBag.stub
@@ -12,5 +12,10 @@ class HeaderBag implements \IteratorAggregate
 	 * @phpstan-return \Traversable<string, string|string[]>
 	 */
 	public function getIterator() {}
+	
+	/**
+	 * @return ($key is null ? array<string, array<int, string|null>> : array<int, string|null>)
+	 */
+	public function all(string $key = null) {}
 
 }


VincentLanglet (approved):

ondrejmirtes (changes requested):
1) I'd say there are more `*Bag` classes with the same behaviour, can you please update stubs for all of them?
2) I'd say we should be able to remove this extension altogether with the help of conditional return types: https://github.com/phpstan/phpstan-symfony/blob/1.2.x/src/Type/Symfony/InputBagDynamicReturnTypeExtension.php

Seldaek:
Yeah, if someone else has time to take this over to do a more comprehensive fix please feel free. I don't know when I'll get to it otherwise. This isn't very high on my list.

VincentLanglet:
For the record, the conditional type was accepted in Symfony 6.3, cf https://github.com/symfony/symfony/pull/49007

Seldaek:
Thanks

PR #283, opened 2022-06-10
Type array in request methods

VincentLanglet:
Same changes than introduced in https://github.com/symfony/symfony/pull/46573
But the pr will only be effective for Symfony 6.2 so this will solve issues for Symfony 4.4+


ondrejmirtes:
Thank you.

PR #287, opened 2022-07-05
Fix exception thrown by Command::run

VincentLanglet:
Command::run is doing
```
try {
     $input->bind($this->definition);
} catch (ExceptionInterface $e) {
     if (!$this->ignoreValidationErrors) {
          throw $e;
    }
}
```
And the bind method is only throwing `ExceptionInterface` so the run method is supposed to thrown only `ExceptionInterface` and not `\Exception` which is too generic.

This avoid error with static analysis when we ignore some kind of exception thrown.

### DIFF
--- a/extension.neon
+++ b/extension.neon
@@ -21,7 +21,9 @@ parameters:
 		- stubs/Symfony/Bundle/FrameworkBundle/Test/KernelTestCase.stub
 		- stubs/Symfony/Bundle/FrameworkBundle/Test/TestContainer.stub
 		- stubs/Symfony/Component/Console/Command.stub
+		- stubs/Symfony/Component/Console/Exception/ExceptionInterface.stub
 		- stubs/Symfony/Component/Console/Helper/HelperInterface.stub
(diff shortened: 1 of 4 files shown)


VincentLanglet:
Friendly ping @ondrejmirtes if you have time to take a look

ondrejmirtes:
I have around 70 unread emails in my inbox after vacation, I'm going from the oldest thread, if you ping here, it means I'm gonna look at it last 😊

VincentLanglet:
> I have around 70 unread emails in my inbox after vacation, I'm going from the oldest thread, if you ping here, it means I'm gonna look at it last 😊

Oh ok.
Do you getting an email for every created PR/opened issues on every phpstan repository ?

Have good vacation then

ondrejmirtes:
Yes, and every comment and every push too 😊

VincentLanglet:
Friendly ping, in case the notification got lost @ondrejmirtes ^^'

VincentLanglet:
I'll close this since it was released in Symfony 4.4, 5.4 and 6.x.
People just have to use the latest versions.

PR #288, opened 2022-07-10
Add generics to cache item

ossinkine:
Add type of cached items to improve type checking

### DIFF
--- a/extension.neon
+++ b/extension.neon
@@ -13,6 +13,7 @@ parameters:
 		consoleApplicationLoader: null
 	stubFiles:
 		- stubs/Psr/Cache/CacheItemInterface.stub
+		- stubs/Psr/Cache/CacheItemPoolInterface.stub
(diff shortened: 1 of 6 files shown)


VincentLanglet:
Hi @ossinkine, adding generics with a phpstan stub is always tricky, because if someone use both psalm and phpstan
writes `CacheItemInterface<Foo>` psalm will report an error because it's not a generic.

Best would be to do the PR directly on php-fig/cache and symfony/symfony repository.
(Symfony now accepts to add generics)

Or at least to provide the same PR on psalm-plugin https://github.com/psalm/psalm-plugin-symfony/tree/master/src/Stubs/common/Contracts/Cache

stof:
As this is about adding generic types for the PSR Cache interface, which is not specific to Symfony, I'm not sure this should be done by the phpstan-symfony extension (and this should probably be discussed with FIG first to try to get them upstream)

ondrejmirtes:
Agreed.

PR #289, opened 2022-07-12
add support for loading SingleCommandApplication

mathroc:
It's already possible to use a single command application, with something like this:

```php
<?php

use Symfony\Component\Console\Application;
use MyApplication;

require __DIR__ . "/../vendor/autoload.php";

$application = new Application();

$application->add(new MyApplication());

return $application;
```

but this makes it even simpler:

```php
<?php

use MyApplication;

require __DIR__ . "/../vendor/autoload.php";

return new MyApplication();
```

### DIFF
--- a/src/Symfony/ConsoleApplicationResolver.php
+++ b/src/Symfony/ConsoleApplicationResolver.php
@@ -7,6 +7,7 @@ use PHPStan\ShouldNotHappenException;
 use PHPStan\Type\ObjectType;
 use Symfony\Component\Console\Application;
 use Symfony\Component\Console\Command\Command;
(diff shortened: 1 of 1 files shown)


ondrejmirtes:
Hi, this PR doesn't allow any new behaviour, it just makes a shortcut possible, right? But since it'd create more documentation and testing burden, I decide not to merge it. Thanks for understanding.

mathroc:
yes that's just a shortcut. 
But it might not be trivial to find out how to handle `SingleCommandApplication`. It's still more documentation but if you want to, I can add an example about it in the readme instead?

ondrejmirtes:
Yes please :)

PR #291, opened 2022-07-15
Add stub for VoterInterface

yguedidi:
To type the `$attributes` with `mixed[]` instead of `array` to be OK until level 9 at least

### DIFF
--- a/extension.neon
+++ b/extension.neon
@@ -40,6 +40,8 @@ parameters:
 		- stubs/Symfony/Component/PropertyAccess/PropertyPathInterface.stub
 		- stubs/Symfony/Component/Security/Acl/Model/AclInterface.stub
 		- stubs/Symfony/Component/Security/Acl/Model/EntryInterface.stub
+		- stubs/Symfony/Component/Security/Core/Authentication/Token/TokenInterface.stub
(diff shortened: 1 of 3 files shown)


herndlm:
Sorry if that's a stupid question, but is that something that could be done in symfony itself too? Or are they not accepting this?

ondrejmirtes:
1) Needs to be registered in stubFiles.
2) What error are you trying to prevent?

yguedidi:
@ondrejmirtes 
1. updated
2. this error: "Method FooVoter::vote() has parameter $attributes with no value type specified in iterable type array." from level 6

yguedidi:
> but is that something that could be done in symfony itself too? Or are they not accepting this?

@stof what do you think about this?

stof:
This might be accepted in Symfony (I would vote for it, but not sure about the rest of the core team).

I think one issue is that PHPStorm has a suggestion (enabled in its default config) to replace it with `array` as being equivalent.

stof:
Note also that this would probably be applied only in the next version of Symfony (as all other type improvements), so the stub might still be useful for the Symfony 5.4 LTS.

yguedidi:
@ondrejmirtes any news here?

ondrejmirtes:
Thank you.

PR #294, opened 2022-07-18
Add stub for options interface

alexander-schranz:
I hope this is correct. First time creating stubs.

Based on discussion in https://github.com/phpstan/phpstan/issues/7650

### DIFF
--- a/stubs/Symfony/Component/OptionsResolver/Options.stub
+++ b/stubs/Symfony/Component/OptionsResolver/Options.stub
@@ -24,7 +24,7 @@ interface Options extends \ArrayAccess, \Countable
 
 	/**
 	 * @template TOffset of key-of<TArray>
-	 * @param TOffset $offset
+	 * @param TOffset|null $offset
 	 * @param TArray[TOffset] $value
 	 *
 	 * @return void
--- a/stubs/Symfony/Component/OptionsResolver/Options.stub
+++ b/stubs/Symfony/Component/OptionsResolver/Options.stub
@@ -1,0 +1,41 @@
+<?php
+
+namespace Symfony\Component\OptionsResolver;
+
+/**
+ * @template TArray of array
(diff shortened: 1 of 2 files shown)


stubs/Symfony/Component/OptionsResolver/Options.stub
```diff
@@ -0,0 +1,41 @@
+<?php
+
+namespace Symfony\Component\OptionsResolver;
```
ondrejmirtes, line 3: Needs to be registered in `stubFiles` in `extension.neon`.

Also since you're making `Options` class generic, you need to add it to `parameters.featureToggles.skipCheckGenericClasses`, similarly to here: https://github.com/phpstan/phpstan-src/blob/74228e6f930c9a0e6ca9461557a6d40b92cf4e91/conf/config.neon#L23-L33 (this list will be purged with the next major version).

ondrejmirtes (changes requested):

ondrejmirtes:
The failures are my bad, with strict-rules these problems are reported: https://phpstan.org/r/2fb146c6-12dc-4a17-bb28-e13d93e820ea

ondrejmirtes:
Reported: https://github.com/phpstan/phpstan/issues/7652

stubs/Symfony/Component/OptionsResolver/Options.stub
```diff
@@ -0,0 +1,41 @@
...
+	/**
+	 * @template TOffset of key-of<TArray>
+	 * @param TOffset $offset
+	 * @return TArray[TOffset]
+	 */
+	public function offsetGet($offset);
+
+	/**
+	 * @template TOffset of key-of<TArray>
+	 * @param TOffset $offset
```
rvanvelzen:
```suggestion
	 * @param TOffset|null $offset
```

ondrejmirtes (approved):

ondrejmirtes:
Perfect, thank you!

PR #296, opened 2022-07-20
Remove Stringable from the `OutputInterface::write()` signature, it's actually not supported

fluffycondor:
I did made a mistake to include Stringable in the OutputInterface signature in #293.
While it works with Symfony 5 because it's typehinted via phpdoc, it fails with Symfony 6 with native type hints.

### DIFF
--- a/extension.neon
+++ b/extension.neon
@@ -15,7 +15,6 @@ parameters:
 		skipCheckGenericClasses:
 			- Symfony\Component\OptionsResolver\Options
 	stubFiles:
-		- stubs/Php/Stringable.stub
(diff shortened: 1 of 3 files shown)


ondrejmirtes:
>  it fails with Symfony 6 with native type hints

What's the error?

fluffycondor:
https://github.com/symfony/console/blob/6.1/Output/Output.php#L138
I guess passing a `Stringable` to `string|iterable` from a context with `declare(strict_types=1);` would result in `TypeError`: https://3v4l.org/QmRqK#v8.1.8

ondrejmirtes:
Thank you.

PR #298, opened 2022-08-09
Add documentation how to ignore container service checks for tests directory

alexander-schranz:


VincentLanglet:
Hi, 2.x branch should be target instead.

Also I don't think this add a lot of value now the original https://phpstan.org/ website is well documented about how to ignore errors. (See https://phpstan.org/user-guide/ignoring-errors)

PR #302, opened 2022-08-18
Add @throws annotation to DataTransformerInterface

VincentLanglet:
This seems to be an issue similar to the lost @phpstan-type in stubs

Since there is no `@throws` in the stub, Phpstan consider that no exception are thrown, until using the real interface definition. I would have expect
- Use @throw void in the stub to override the interface
- No @throw = default definition from the interface

This would be fixed this way by the whole changed made in https://github.com/phpstan/phpstan-src/pull/1603

Until this is globally fixed (or not if considered as normal behavior...), I made a PR to fix the stub here @ondrejmirtes


ondrejmirtes:
Thank you.

PR #300, opened 2022-08-15
Add support for negatable input options

eliashaeussler:
Since symfony/console 5.3 (https://github.com/symfony/symfony/pull/39642), input options can be marked as "negatable". The appropriate return type when calling `$input->getOption()` can be either a boolean or `null`. With this change, support for return types of such negatable input options is added.

PR #305, opened 2022-08-26
Skip check for negatable input options on symfony/console < 5.3

eliashaeussler:
This fixes a regression introduced with #300 where the `InputOption::isNegatable()` method is not available for `symfony/console` < 5.3.0.


ondrejmirtes:
Thank you, I wanted to do exactly this :) I'm gonna release it immediately.

PR #309, opened 2022-11-04
Fix Envelope::all() return type

enumag:
I noticed that my fix from #308 wasn't working so I looked deeper into it. Trying a better fix now.

### DIFF
--- a/src/Type/Symfony/EnvelopeReturnTypeExtension.php
+++ b/src/Type/Symfony/EnvelopeReturnTypeExtension.php
@@ -5,10 +5,12 @@ namespace PHPStan\Type\Symfony;
 use PhpParser\Node\Expr\MethodCall;
 use PHPStan\Analyser\Scope;
 use PHPStan\Reflection\MethodReflection;
+use PHPStan\Type\Accessory\AccessoryArrayListType;
 use PHPStan\Type\ArrayType;
 use PHPStan\Type\Constant\ConstantStringType;
 use PHPStan\Type\DynamicMethodReturnTypeExtension;
-use PHPStan\Type\MixedType;
+use PHPStan\Type\Generic\GenericClassStringType;
+use PHPStan\Type\IntegerType;
 use PHPStan\Type\ObjectType;
 use PHPStan\Type\Type;
 use function count;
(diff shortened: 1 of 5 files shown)


src/Type/Symfony/EnvelopeReturnTypeExtension.php
```diff
@@ -33,15 +35,15 @@ public function getTypeFromMethodCall(
...
 		}
 
 		$argType = $scope->getType($methodCall->getArgs()[0]->value);
 		if (!$argType instanceof ConstantStringType) {
-			return new ArrayType(new MixedType(), new ObjectType('Symfony\Component\Messenger\Stamp\StampInterface'));
+			return new ArrayType(new IntegerType(), new ObjectType('Symfony\Component\Messenger\Stamp\StampInterface'));
 		}
 
-		return new ArrayType(new MixedType(), new ObjectType($argType->getValue()));
+		return new ArrayType(new IntegerType(), new ObjectType($argType->getValue()));
```
enumag: Hmm... how can I tell PHPStan here that it's a `list<>`? :thinking:

src/Type/Symfony/EnvelopeReturnTypeExtension.php
```diff
@@ -33,15 +36,15 @@ public function getTypeFromMethodCall(
 	): Type
 	{
 		if (count($methodCall->getArgs()) === 0) {
-			return new ArrayType(new MixedType(), new ArrayType(new MixedType(), new ObjectType('Symfony\Component\Messenger\Stamp\StampInterface')));
+			return new ArrayType(new ClassStringType(), new ArrayType(new IntegerType(), new ObjectType('Symfony\Component\Messenger\Stamp\StampInterface')));
```
enumag: And here... how do I say that the ClassStringType is actually a `class-string<StampInterface>`?

ondrejmirtes: By checking out the table here https://phpstan.org/developing-extensions/type-system or API reference here https://apiref.phpstan.org/1.9.x/PHPStan.Type.Type.html

src/Type/Symfony/EnvelopeReturnTypeExtension.php
ondrejmirtes: The test needs to run with bleedingEdge config, put it in `getAdditionalConfigFiles`.

```
'phar://' . __DIR__ . '/../../../vendor/phpstan/phpstan/phpstan.phar/conf/bleedingEdge.neon'
```

enumag: But even then this says `array<int, ...>` instead of `list<...>` - and these do have different meaning. I don't think PHPStan should assume that every array with integer keys is a list.

I could not find any ListType or ListArrayType or anything like that.

enumag: Yeah, as expected bleeding edge config didn't change anything.

src/Type/Symfony/EnvelopeReturnTypeExtension.php
enumag: GenericClassStringType... okay, solved

src/Type/Symfony/EnvelopeReturnTypeExtension.php
ondrejmirtes: You need to call AccessoryArrayListType::intersect() on the type you want to become a list.

enumag: I found AccessoryArrayListType but it doesn't take an argument to specify the value type so I don't think that's what I need here either.

ondrejmirtes: Just look how it's used in phpstan-src.

enumag: Ah okay. This works when combined with bleeding edge.

```php
AccessoryArrayListType::intersectWith(new ArrayType(new IntegerType(), new ObjectType('Symfony\Component\Messenger\Stamp\StampInterface'))),
```

enumag:
Hmm... except that `AccessoryArrayListType` doesn't exist in PHPStan 1.8. How can I solve that? `list<>` was available even before phpstan 1.9.

Also some tests are now failing on this:
```
Error: Calling PHPStan\Type\Accessory\AccessoryArrayListType::intersectWith() is not covered by backward compatibility promise. The method might change in a minor PHPStan version.
```

ondrejmirtes:
Require PHPStan 1.9

enumag:
Okay... I'll go and assume that the method should be covered by compatibility promise since you recommended it's usage... see https://github.com/phpstan/phpstan-src/pull/1967.

ondrejmirtes (changes requested):
You can require PHPStan ^1.9.1 now.

ondrejmirtes:
Thank you.

PR #310, opened 2022-11-12
Fallback to default helper set when no application loader is provided

acrobat:
I'm using the phpstan-symfony extension in a library so I'm not able to register the standard console application loader (as you would in a regular project)

So I've added a fallback to the default helper set symfony will register during setup of the console application. 

See: https://github.com/symfony/symfony/blob/ea9ed6c821f76dfc5cf0639ebd1a23db971dc6e4/src/Symfony/Component/Console/Application.php#L1096-L1107

Most of the `getHelper` cases can be handled by this default set. When someone has custom helpers they should use the `consoleApplicationLoader` option.


ondrejmirtes:
Please explain what effect does this have on real-world applications, thanks.

> I'm using the phpstan-symfony extension in a library 

You can still create a Console Application just for the purpose of static analysis with PHPStan.

PR #317, opened 2022-12-18
Add stub for AbstractType

gndk:
This PR adds stubs for `Symfony\Component\Form\AbstractType` and `Symfony\Component\Form\AbstractTypeExtension`.

After upgrading to Psalm 5, it requires `@template-extends AbstractType<int>` annotation when extending `AbstractType`, but currently phpstan complains about this. 

Adding these stubs creates parity with [psalm/psalm-plugin-symfony](https://github.com/psalm/psalm-plugin-symfony/tree/5.x/src/Stubs/common/Component/Form) regarding those classes.

### DIFF
--- a/extension.neon
+++ b/extension.neon
@@ -13,6 +13,10 @@ parameters:
 		consoleApplicationLoader: null
 	featureToggles:
 		skipCheckGenericClasses:
+			- Symfony\Component\Form\AbstractType
+			- Symfony\Component\Form\AbstractTypeExtension
+			- Symfony\Component\Form\FormTypeExtensionInterface
+			- Symfony\Component\Form\FormTypeInterface
 			- Symfony\Component\OptionsResolver\Options
 	stubFiles:
 		- stubs/Psr/Cache/CacheItemInterface.stub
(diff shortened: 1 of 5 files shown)


stof:
what is expected to be represented by that template type ?

gndk:
The stubs in the psalm plugin were originally added in this PR https://github.com/psalm/psalm-plugin-symfony/pull/138, with some discussion here https://github.com/psalm/psalm-plugin-symfony/issues/60.

gndk:
There is also discussion around this current issue after upgrading to psalm 5 here https://github.com/psalm/psalm-plugin-symfony/issues/294

VincentLanglet (changes requested):
Those stubs are useless for PHPStan and even more they will just add more pain for PHPStan-Symfony user because it will require to fill a template
```
/**
 * @extends AbstractType<ICanWriteEverythingIWant>
 */
class MyType extends AbstractType
{
}
```
With no benefit at all.

@stof Those stubs were added to psalm-symfony with only one benefit, this one:
https://github.com/psalm/psalm-plugin-symfony/blob/4.x/src/Stubs/common/Component/Form/FormView.stubphp#L16
but the syntax `array{value: ?T, attr: array<array-key, mixed>}&array<string, mixed>`
is not supported by PHPStan.

So there is absolutely no reason to add such template.

Also the symfony-core was not really into those generics https://github.com/symfony/symfony/pull/40783#issuecomment-953098461

If you want to use both psalm and phpstan, I assume you can write
```
/** 
 * @psalm-param FormView<Foo>
 * @phpstan-param FormView
 */
```

PR #318, opened 2022-12-22
Add generics for PasswordUpgraderInterface

VincentLanglet:
This is https://github.com/symfony/symfony/pull/48750 but for older Symfony version.
I'll wait for the Symfony merge this.

### DIFF
--- a/extension.neon
+++ b/extension.neon
@@ -14,6 +14,7 @@ parameters:
 	featureToggles:
 		skipCheckGenericClasses:
 			- Symfony\Component\OptionsResolver\Options
+			- Symfony\Component\Security\Core\User\PasswordUpgraderInterface
 	stubFiles:
 		- stubs/Psr/Cache/CacheItemInterface.stub
 		- stubs/Symfony/Bundle/FrameworkBundle/KernelBrowser.stub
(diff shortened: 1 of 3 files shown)


VincentLanglet:
https://github.com/symfony/symfony/pull/48750 was merged so this is ready to be merged @ondrejmirtes

ondrejmirtes:
I can only merge this if the new generic class is added to `skipCheckGenericClasses` in `extension.neon`.

VincentLanglet:
> skipCheckGenericClasses

Done !
What's the idea behind ? Is it for BC purpose ?

ondrejmirtes:
Yes, I'm gonna empty it in the next major version.

ondrejmirtes:
Thank you!

VincentLanglet:
> Yes, I'm gonna empty it in the next major version.

I assume that if you use bleedingEdge, skipCheckGenericClasses has no effect ?

ondrejmirtes:
There's https://github.com/phpstan/phpstan-src/blob/3408a97088d6af676114814fa83455b11ac4da76/conf/bleedingEdge.neon#L4 but I think that only applies to https://github.com/phpstan/phpstan-src/blob/3408a97088d6af676114814fa83455b11ac4da76/conf/config.neon#L23-L33 because phpstan-symfony config is applied later. Anyway, you can verify this yourself :)

PR #324, opened 2023-01-14
Revert wrong typehint for getSession return type

### LINKED ISSUE #323: Session/SessionInterface inconsistency
Hi @ondrejmirtes,

After the merge of https://github.com/phpstan/phpstan-symfony/commit/012305dab7a08f4e1a7158e01d7b4bccfbc6fa31
(cf https://github.com/phpstan/phpstan-symfony/pull/233)

The return type of `Request::getSession` is Session instead of SessionInterface.
This was not really true be could be considered as a good help for developers since
- it was almost always true
- there was no easy way to check for the `getFlashBag` method.

1) This is not the case after the merge of https://github.com/symfony/symfony/commit/c7b5802f8ae1bd901e2506bc9f2b4b72522eb817 in Symfony 6.2.
Now, it's possible to do a `$request->getSession() instanceof FlashBagAwareSessionInterface` check.
And, because of the stub, currently the stub is reported as always true which is not true...

2) Currently, `RequestStack::getSession` is returning `SessionInterface` when `Request::getSession()` is returning `Session`.
This should be fixed by reverting the stub for request, or adding a stub for RequestStack.

3) The Request stub is loaded dynamically with the following extension https://github.com/phpstan/phpstan-symfony/blob/1.2.x/src/Symfony/InputBagStubFilesExtension.php#L19
which means that 
- For user of SF <5.1, `$request::getSession` returns a SessionInterface.
- For user of SF >=5.1, `$request::getSession` returns a Session.

Not sure this was expected

I'd like to improve the situation, what's your point of view about this ?
The easiest thing to do would be at least to introduce a stub for RequestStack, but there is maybe something better to do ?

VincentLanglet:
TLDR: This would be one way to closes https://github.com/phpstan/phpstan-symfony/issues/323, and IMHO it's the best one because
- it's correct in 100% of the time for Symfony users and promotes to right correct code.
- People who doesn't care about the Session/SessionInterface can ignore the error.
- People who care about the Session/SessionInterface are not dealing with wrong errors.

-----------------

This type hint was introduced by
https://github.com/phpstan/phpstan-symfony/commit/012305dab7a08f4e1a7158e01d7b4bccfbc6fa31
which was done to replace the original PR
https://github.com/phpstan/phpstan-symfony/pull/233 cc @ruudk 

As explained in https://github.com/symfony/symfony/issues/39222#issuecomment-735469831
or https://github.com/symfony/symfony/issues/41765#issuecomment-870268556:
- This is not true that `getSession` always return a Session
- People should rely on a `instanceof` checks
This is even easier now that the `FlashBagAwareSessionInterface` was introduced in https://github.com/symfony/symfony/issues/41765.

It was said by @ruudk 
> For a small set of users it could mean that phpstan now falsely allows getFlashBag on the request session, because they have a custom session implementation. They prob didn't call the flashbag anyway, as it doesn't work.

But this is not the real problem of the stub.
The real issue is that if you use service decoration, or write correctly your calls to support multiple sessions, like
```
$session = $request->getSession();
if ($session instanceof FlashBagAwareSessionInterface) {
     $session->getFlashBag()->addMessage('Foo');
} elseif ($session instanceof MyCustomSessionInterface) {
     $session->doThings();
} else {
     // ...
}
```
You ended up with false positive from phpstan about `always true instanceof` or `always false instanceof`.
This is way better to have a "false negative" (`getFlashBag() doesn't exist in SessionInterface) than introducing false positive.

If people doesn't want to write correct code about Symfony session, they should ignore the error 
```
ignoreErros:
     - '#Call to an undefined method Symfony\\Component\\HttpFoundation\\Session\\SessionInterface::getFlashBag\(\)#'
```
rather than introducing a wrong stub.

Closes https://github.com/phpstan/phpstan-symfony/issues/323

### DIFF
--- a/stubs/Symfony/Component/HttpFoundation/Request.stub
+++ b/stubs/Symfony/Component/HttpFoundation/Request.stub
@@ -27,11 +27,6 @@ class Request
 	public $cookies;
 
     /**
-     * @return \Symfony\Component\HttpFoundation\Session\Session
-     */
-    public function getSession();
-
-    /**
      * @return string[]
      */
     public static function getTrustedProxies(): array;
--- a/tests/Type/Symfony/data/request_get_session.php
+++ b/tests/Type/Symfony/data/request_get_session.php
@@ -1,14 +1,14 @@
 <?php declare(strict_types = 1);
 
-use Symfony\Component\HttpFoundation\Session\Session;
+use Symfony\Component\HttpFoundation\Session\SessionInterface;
 use function PHPStan\Testing\assertType;
 
 /** @var \Symfony\Component\HttpFoundation\Request $request */
 $request = doRequest();
 
 $session1 = $request->getSession();
-assertType(Session::class, $request->getSession());
+assertType(SessionInterface::class, $request->getSession());
 
 if ($request->hasSession()) {
-	assertType(Session::class, $request->getSession());
+	assertType(SessionInterface::class, $request->getSession());
 }
--- a/tests/Type/Symfony/data/request_get_session_null.php
+++ b/tests/Type/Symfony/data/request_get_session_null.php
@@ -1,14 +1,14 @@
 <?php declare(strict_types = 1);
 
-use Symfony\Component\HttpFoundation\Session\Session;
+use Symfony\Component\HttpFoundation\Session\SessionInterface;
 use function PHPStan\Testing\assertType;
 
 /** @var \Symfony\Component\HttpFoundation\Request $request */
 $request = doRequest();
 
 $session1 = $request->getSession();
-assertType(Session::class . '|null', $request->getSession());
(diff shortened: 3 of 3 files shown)


ruudk:
I looked at it again, I think that retrieving it from the Request is coming from the early days. Now with ArgumentValueResolvers this can be done simpler:

```php
public function myAction(Session $session) : array
{
    $session->getFlashBag()->add('success', 'Hi!');
}
```
or
```php
public function myAction(FlashBagAwareSessionInterface $session) : array
{
    $session->getFlashBag()->add('success', 'Hi!');
}
```

This code doesn't require any `instanceof` checks. 

So I'm 👍 with shipping this PR.

ondrejmirtes:
Thank you both!

Issue #322, opened 2023-01-09
Options stub doesn't work as expected


VincentLanglet:
Related to https://github.com/phpstan/phpstan/issues/7650
and PR https://github.com/phpstan/phpstan-symfony/pull/294
cc @alexander-schranz and @ondrejmirtes 

I saw the following example https://phpstan.org/r/95a7bcea-3198-4297-8770-823861adacfb
(btw now it's full of error: https://phpstan.org/r/339c6e0b-c988-4a01-b203-1047ec4f71ee)

I wrote the following code
```
/**
     * @param Options<array{
     *     page: PageInterface|null,
     *     site: SiteInterface|null,
     *     choice_translation_domain: bool,
     *     filter_choice: array{current_page: false, request_method: string, dynamic: bool, hierarchy: string},
     * }> $options
     *
     * @return array<PageInterface>
     */
    private function getChoices(Options $options): array
    {
        \PHPStan\dumpType($options['site']);
```
And I'm getting 
```
Dumped type: array{current_page: false, request_method: string, dynamic: bool, hierarchy:                                                                          
         string}|bool|Sonata\PageBundle\Model\PageInterface|Sonata\PageBundle\Model\SiteInterface|null
```

But I cannot reproduce the issue in the phpstan reproducer.

If you look at this PR: https://github.com/sonata-project/SonataPageBundle/pull/1654,
you'll see that
- When using the Options from Symfony, I get the wrong inference.
- When using my Options, it works fine.

Am I missing something ?

VincentLanglet:
I think it's related to https://github.com/phpstan/phpstan/issues/8549

Since the example given https://phpstan.org/r/95a7bcea-3198-4297-8770-823861adacfb
is now throwing errors: https://phpstan.org/r/c121a4ee-398d-4fa5-88d9-313f2d21d04c

@ondrejmirtes Until the bug is fixed shouldn't the stub be disabled ?
Or at least is there a way to keep this line https://github.com/phpstan/phpstan-symfony/blob/1.2.x/extension.neon#L16
with bleedingEdge ?

VincentLanglet:
@alexander-schranz Does the Options stub works for you ?

PR #326, opened 2023-01-19
Introduce Voter stub

VincentLanglet:
Does such a stub require a test ?

### DIFF
--- a/stubs/Symfony/Component/Security/Core/Authorization/Voter/Voter.stub
+++ b/stubs/Symfony/Component/Security/Core/Authorization/Voter/Voter.stub
@@ -1,0 +1,35 @@
+<?php
+
+namespace Symfony\Component\Security\Core\Authorization\Voter;
+
+use Symfony\Component\Security\Core\Authentication\Token\TokenInterface;
+
+/**
+ * @template TAttribute of string
+ * @template TSubject of mixed
+ */
+abstract class Voter implements VoterInterface
+{
+    /**
+     * Determines if the attribute and subject are supported by this voter.
+     *
+     * @param mixed $subject
+     *
+	 * @phpstan-assert-if-true TSubject $subject
+	 * @phpstan-assert-if-true TAttribute $attribute
+     *
+     * @return bool
+     */
+    abstract protected function supports(string $attribute, $subject);
+
+    /**
+     * Perform a single access check operation on a given attribute, subject and token.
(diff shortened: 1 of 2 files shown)


ondrejmirtes:
I'm not sure that everyone who `extends Voter` wants this functionality. In case they do, try introducing these generics into Symfony directly.

VincentLanglet:
> I'm not sure that everyone who `extends Voter` wants this functionality.

Same could be said for almost every stub I would say.
For instance I get lot of trouble with this stub https://github.com/phpstan/phpstan-symfony/issues/322

I don't consider this stub as an hack or a wrong stub.
And I thought this was one of the purpose of `skipCheckGenericClasses`, people can skip this stub.
Or just `Voter<string, mixed>` will do nothing.

> In case they do, try introducing these generics into Symfony directly.

Symfony is a kinda slow adopter about new annotations as shown in https://github.com/symfony/symfony/pull/46613
Also, they only accepts such things on new branch, which means it would be only available on Sf 6.3.

But I can start by this to get feedback and decide after for phpstan-symfony, sure. 👍 
Edit: opened https://github.com/symfony/symfony/pull/49033

VincentLanglet:
> I'm not sure that everyone who extends Voter wants this functionality. In case they do, try introducing these generics into Symfony directly.

Hi @ondrejmirtes, the template was accepted by symfony in 6.3 https://github.com/symfony/symfony/pull/49033
Do we merge this to add the template in older symfony version ?

stubs/Symfony/Component/Security/Core/Authorization/Voter/Voter.stub
```diff
@@ -0,0 +1,35 @@
...
+ * @template TSubject of mixed
+ */
+abstract class Voter implements VoterInterface
+{
+    /**
+     * Determines if the attribute and subject are supported by this voter.
+     *
+     * @param mixed $subject
+     *
+	 * @phpstan-assert-if-true TSubject $subject
```
ondrejmirtes: Wrong formatting

ondrejmirtes (changes requested):

stubs/Symfony/Component/Security/Core/Authorization/Voter/Voter.stub
VincentLanglet: Nice catch, fixed.

ondrejmirtes (approved):

ondrejmirtes:
Thank you.

PR #327, opened 2023-01-31
Extend NormalizerInterface::normalize return type, add @throws to NormalizerInterface and DenormalizerInterface

hxv:
Two issues in Serializer component stubs:
- `\Symfony\Component\Serializer\Normalizer\NormalizerInterface::normalize()` can return `\ArrayObject` ([source](https://github.com/symfony/serializer/blob/v6.2.5/Normalizer/NormalizerInterface.php#L31)) – type is missing in `@return` annotation.
- Both `\Symfony\Component\Serializer\Normalizer\NormalizerInterface::normalize()` ([source](https://github.com/symfony/serializer/blob/v6.2.5/Normalizer/NormalizerInterface.php#L33-L37)) and `\Symfony\Component\Serializer\Normalizer\DenormalizerInterface::denormalize()` ([source](https://github.com/symfony/serializer/blob/v6.2.5/Normalizer/DenormalizerInterface.php#L39-L45)) throw exceptions – `@throws` annotations are missing from stubs completely. This looks like problem from #302 in another component.

### DIFF
--- a/extension.neon
+++ b/extension.neon
@@ -58,6 +58,14 @@ parameters:
 		- stubs/Symfony/Component/Serializer/Encoder/ContextAwareDecoderInterface.stub
 		- stubs/Symfony/Component/Serializer/Encoder/DecoderInterface.stub
 		- stubs/Symfony/Component/Serializer/Encoder/EncoderInterface.stub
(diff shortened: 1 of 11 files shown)


ondrejmirtes:
Thank you!

PR #329, opened 2023-02-04
Add missing @throws in stubs.

hxv:
I've compared all stub files with Symfony's source files, and found some missing `@throws` tags:
- in `Symfony\Contracts\Cache\CacheInterface::get()` ([source](https://github.com/symfony/cache-contracts/blob/v1.1.6/CacheInterface.php#L43))
- in `Symfony\Component\Console\Command\Command::getHelper()` ([source](https://github.com/symfony/console/blob/v6.2.5/Command/Command.php#L686-L687))
- in `Symfony\Component\DependencyInjection\Extension\ExtensionInterface::load()` ([source](https://github.com/symfony/dependency-injection/blob/v6.2.5/Extension/ExtensionInterface.php#L28))
- in `Symfony\Component\Serializer\Encoder\EncoderInterface::encode()` ([source](https://github.com/symfony/serializer/blob/v6.2.5/Encoder/EncoderInterface.php#L28))
- in `Symfony\Component\Serializer\Encoder\DecoderInterface::decode()` ([source](https://github.com/symfony/serializer/blob/v6.2.5/Encoder/DecoderInterface.php#L35))

Exceptions should now match.

### DIFF
--- a/extension.neon
+++ b/extension.neon
@@ -17,11 +17,16 @@ parameters:
 			- Symfony\Component\Security\Core\Authorization\Voter\Voter
 			- Symfony\Component\Security\Core\User\PasswordUpgraderInterface
 	stubFiles:
+		- stubs/Psr/Cache/CacheException.stub
 		- stubs/Psr/Cache/CacheItemInterface.stub
+		- stubs/Psr/Cache/InvalidArgumentException.stub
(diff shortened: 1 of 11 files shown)


ondrejmirtes:
Thank you.

PR #334, opened 2023-02-17
Add InvalidLazyCommandRule

jmleroux:
My first try to add a rule to detect not lazy commands.

See here for more info about lazy commands: https://symfony.com/blog/new-in-symfony-5-3-lazy-command-description

Do not hesitate to tel me if I have to add more info for this kind of PR.


ondrejmirtes:
Thank you, but this isn't a good fit for phpstan-symfony. Symfony doesn't require all commands to be lazy, so this rule would be annoying for mainstream use.

I encourage you to publish a custom package with this rule. Thank you for your understanding.

PR #337, opened 2023-02-20
feat: Support specifying type of data_class in forms

siketyan:
You can now specify what type of data_class the FormType have as the generic parameter of FormTypeInterface or AbstractType.

Example:

```php
/**
 * @extends AbstractType<DataClass>
 */
class DataClassType extends AbstractType
```

Also, you can validate mismatch between data_class in FormType and the initial data:

```php
$form = $this->formFactory->create(DataClassType::class, new DataClass()); // OK
$form = $this->formFactory->create(DataClassType::class, new AnotherClass()); // ERROR!
```

Then you can get appropriate type of the data:

```php
$data = $form->getData();
\PHPStan\dumpType($data): // Dumped type: DataClass
```


stubs/Symfony/Component/Form/FormFactoryInterface.stub
```diff
@@ -0,0 +1,36 @@
...
+	public function create(string $type = FormType::class, $data = null, array $options = []): FormInterface;
+
+	/**
+	 * @template TFormType of FormTypeInterface<TData>
+	 * @template TData
+	 *
+	 * @param class-string<TFormType> $type
+	 * @param TData $data
+	 * @param array<string, mixed> $options
+     *
```
ondrejmirtes: Wrong indentation

tests/Stubs/Symfony/Component/Form/DataClassType.php
```diff
@@ -0,0 +1,34 @@
...
+use Symfony\Component\Form\AbstractType;
+use Symfony\Component\Form\Extension\Core\Type\NumberType;
+use Symfony\Component\Form\Extension\Core\Type\TextType;
+use Symfony\Component\Form\FormBuilderInterface;
+use Symfony\Component\OptionsResolver\OptionsResolver;
+
+/**
+ * @extends AbstractType<DataClass>
+ */
+class DataClassType extends AbstractType
```
ondrejmirtes: Not sure what these tests are supposed to do, they're not referenced from anywhere.

stubs/Symfony/Component/Form/FormFactoryInterface.stub
```diff
@@ -0,0 +1,36 @@
...
+{
+	/**
+	 * @template TFormType of FormTypeInterface<TData>
+	 * @template TData
+	 *
+	 * @param class-string<TFormType> $type
+	 * @param TData $data
+	 * @param array<string, mixed> $options
+	 *
+	 * @phpstan-return ($data is null ? FormInterface<null|TData> : FormInterface<TData>)
```
ondrejmirtes, line 17: These complex generics should be tested with TypeInferenceTestCase. To make sure they behave like you expect them to.

stubs/Symfony/Component/Form/AbstractType.stub
```diff
@@ -0,0 +1,12 @@
+<?php
+
+namespace Symfony\Component\Form;
+
+/**
+ * @template TData
```
ondrejmirtes, line 6: Every time a new class is made generic, the class name should be added here: https://github.com/phpstan/phpstan-symfony/blob/1da7bf450c6b351fec08ca0aa97298473d4f6ab3/extension.neon#L14-L18 (to make sure users are not bothered by "Function foo() has parameter $hw with generic class HelloWorld but does not specify its types: T" errors.

This list will be made empty with the next major version.

ondrejmirtes (changes requested):
You've given a compelling list of arguments! Looking forward to merge this.

stubs/Symfony/Component/Form/FormInterface.stub
```diff
@@ -3,10 +3,15 @@
...
+ * @template TData
+ *
+ * @extends \ArrayAccess<string, \Symfony\Component\Form\FormInterface<mixed>>
+ * @extends \Traversable<string, \Symfony\Component\Form\FormInterface<mixed>>
  */
 interface FormInterface extends \ArrayAccess, \Traversable, \Countable
 {
-
+    /**
+     * @return TData
```
VincentLanglet, line 21: I didn't look where is the truth, but the Psalm template returns `?TData`
https://github.com/psalm/psalm-plugin-symfony/blob/5.x/src/Stubs/common/Component/Form/FormInterface.stubphp#L12

tests/Stubs/Symfony/Component/Form/DataClassType.php
siketyan: It's used from here: https://github.com/siketyan/phpstan-symfony/blob/a18b756621163b0040a3c3f5a58a2f24b3652f48/tests/Stubs/Symfony/Component/Form/FormFactoryAwareClass.php#L20.
Although it is for testing purpose to confirm the FormFactoryInterface accepts `DataClass` for `DataClassType` form type.

ondrejmirtes: That’s not how testing works here 😊 Please see https://phpstan.org/developing-extensions/testing to know how to test it.

stubs/Symfony/Component/Form/FormInterface.stub
siketyan: I think it will be null only if we omit a initial value of the form.
https://github.com/siketyan/phpstan-symfony/blob/88da905c8a29f791fa9ee78ec5f83f2980f8e809/stubs/Symfony/Component/Form/FormFactoryInterface.stub#L17

stubs/Symfony/Component/Form/AbstractType.stub
siketyan: Added in [da2cb98](https://github.com/phpstan/phpstan-symfony/pull/337/commits/da2cb981eb40944ba69a93595b7f4e3dedd2b40b)

stubs/Symfony/Component/Form/FormFactoryInterface.stub
siketyan: Fixed in [88da905](https://github.com/phpstan/phpstan-symfony/pull/337/commits/88da905c8a29f791fa9ee78ec5f83f2980f8e809)

tests/Stubs/Symfony/Component/Form/DataClassType.php
siketyan: How about this? [dcf9b4f](https://github.com/phpstan/phpstan-symfony/pull/337/commits/dcf9b4ff13233b72e7c6c4f8eb86abfe87601dc5)

stubs/Symfony/Component/Form/FormInterface.stub
VincentLanglet: I see. So it means psalm should improve his typing ^^

ondrejmirtes (approved):

ondrejmirtes:
Thank you!

blankse:
@siketyan @ondrejmirtes After update to 1.3.0 I get now following errors:
```
Parameter $form of method App\Form\Type\TypeaheadType::buildView() has invalid type Symfony\Component\Form\TData.
```
I haven't much experience with the generics. I am only on phpstan level 6 with disabled checkGenericClassInNonGenericObjectType and checkMissingIterableValueType.

The form types have no data class. So I tried to add following phpdocs to the classes, but it doesn't help:
```
/**
 * @extends AbstractType<null>
 */
```
```
/**
 * @extends AbstractType<mixed>
 */
```

Is there a bug on your side or what make I wrong?

ondrejmirtes:
Please show a piece of code for which PHPStan reports this error:

> Parameter $form of method App\Form\Type\TypeaheadType::buildView() has invalid type Symfony\Component\Form\TData.

blankse:
@ondrejmirtes
```
<?php

namespace App\Form\Type;

use Symfony\Component\Form\AbstractType;
use Symfony\Component\Form\Extension\Core\Type\TextType;
use Symfony\Component\Form\FormInterface;
use Symfony\Component\Form\FormView;
use Symfony\Component\OptionsResolver\OptionsResolver;

/**
 * @extends AbstractType<mixed>
 */
class TypeaheadType extends AbstractType
{
    public function configureOptions(OptionsResolver $resolver): void
    {
        $resolver->setDefaults([
            'definitionType' => 'class',
            'definitionId' => null,
            'field' => null,
            'multiple' => false,
            'type' => '',
            'documentId' => null,
        ]);
    }

    public function buildView(FormView $view, FormInterface $form, array $options): void
    {
        if (!$options['definitionId'] || !$options['field']) {
            throw new \InvalidArgumentException('Typeahead: No definitionId or field is set');
        }

        $view->vars = array_replace_recursive($view->vars, [
            'definitionType' => $options['definitionType'],
            'definitionId' => $options['definitionId'],
            'field' => $options['field'],
            'multiple' => $options['multiple'],
            'documentId' => $options['documentId'],
            'attr' => [
                'class' => 'js-typeahead ' . $options['type'],
            ]
        ]);
    }

    public function getParent(): string
    {
        return TextType::class;
    }
}
```

ondrejmirtes:
Fixed: https://github.com/phpstan/phpstan-symfony/commit/7e78605a699d183f5a6936cf91904f4c16ca79b2

There was a weird bug because all the methods are re-defined in AbstractType even if they're in the FormTypeInterface ancestor. So the stub file needs to repeat them too.

siketyan:
Thank you for the quick fix @ondrejmirtes and sorry for my mistake @blankse

blankse:
@ondrejmirtes Thank you for the quick fix 👍🏼

PR #342, opened 2023-03-23
Add stubs for security factories

yguedidi:
ref symfony/symfony#49783

### DIFF
--- a/stubs/Symfony/Bundle/SecurityBundle/DependencyInjection/Security/Factory/AuthenticatorFactoryInterface.stub
+++ b/stubs/Symfony/Bundle/SecurityBundle/DependencyInjection/Security/Factory/AuthenticatorFactoryInterface.stub
@@ -1,0 +1,16 @@
+<?php
+
+namespace Symfony\Bundle\SecurityBundle\DependencyInjection\Security\Factory;
+
+use Symfony\Component\Config\Definition\Builder\NodeDefinition;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+
+interface AuthenticatorFactoryInterface
+{
+    /**
+     * @param array<string, mixed> $config
+     *
+     * @return string|string[]
+     */
+    public function createAuthenticator(ContainerBuilder $container, string $firewallName, array $config, string $userProviderId);
+}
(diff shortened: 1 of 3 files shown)


stubs/Symfony/Bundle/SecurityBundle/DependencyInjection/Security/Factory/AuthenticatorFactoryInterface.stub
```diff
@@ -0,0 +1,16 @@
...
+
+use Symfony\Component\Config\Definition\Builder\NodeDefinition;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+
+interface AuthenticatorFactoryInterface
+{
+    /**
+     * @param array<string, mixed> $config
+     *
+     * @return string|string[]
```
ondrejmirtes, line 13: When does it return a string and when does it return an array?

ondrejmirtes (changes requested):

stubs/Symfony/Bundle/SecurityBundle/DependencyInjection/Security/Factory/AuthenticatorFactoryInterface.stub
yguedidi: It depends on the concrete implementation, for example:

- string: https://github.com/symfony/symfony/blob/6.3/src/Symfony/Bundle/SecurityBundle/DependencyInjection/Security/Factory/AccessTokenFactory.php#L70
- array: https://github.com/symfony/symfony/blob/6.3/src/Symfony/Bundle/SecurityBundle/DependencyInjection/Security/Factory/CustomAuthenticatorFactory.php#L60

stof: Returning a string is treated the same than returning an array containing a single item by the Symfony code calling this factory. It allows to simplify the most common case which only needs to create a single authenticator.

yguedidi: @ondrejmirtes is there anything else I can do?

stof (approved):

ondrejmirtes:
Thank you!

ondrejmirtes:
Thank you!

PR #348, opened 2023-04-21
Set properties autowired with @required as initialized

### LINKED ISSUE #346: checkUninitializedProperties does not consider @required
When enabling PHPStan's `checkUninitializedProperties`, using `@required` to autowire a public property, or through a public method will still mark the property as uninitialized, although Symfony's autowiring will make sure it isn't.

This is on;
phpstan/phpstan 1.10.13
phpstan/phpstan-symfony 1.3.1

raalderink:
Fixes https://github.com/phpstan/phpstan-symfony/issues/346

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -15,7 +15,7 @@
   "require": {
     "php": "^7.2 || ^8.0",
     "ext-simplexml": "*",
-    "phpstan/phpstan": "^1.9.18"
+    "phpstan/phpstan": "^1.10.36"
   },
   "conflict": {
     "symfony/framework-bundle": "<3.0"
@@ -35,7 +35,8 @@
     "symfony/http-foundation": "^5.4 || ^6.1",
     "symfony/messenger": "^5.4",
     "symfony/polyfill-php80": "^1.24",
-    "symfony/serializer": "^5.4"
+    "symfony/serializer": "^5.4",
+    "symfony/service-contracts": "^2.2.0"
   },
   "config": {
     "sort-packages": true
--- a/src/Symfony/RequiredAutowiringExtension.php
+++ b/src/Symfony/RequiredAutowiringExtension.php
@@ -1,0 +1,91 @@
+<?php declare(strict_types = 1);
+
+namespace PHPStan\Symfony;
+
+use PHPStan\Reflection\AdditionalConstructorsExtension;
+use PHPStan\Reflection\ClassReflection;
+use PHPStan\Reflection\Php\PhpPropertyReflection;
+use PHPStan\Reflection\PropertyReflection;
+use PHPStan\Rules\Properties\ReadWritePropertiesExtension;
+use PHPStan\Type\FileTypeMapper;
+use function count;
+
+class RequiredAutowiringExtension implements ReadWritePropertiesExtension, AdditionalConstructorsExtension
+{
+
+	/** @var FileTypeMapper */
+	private $fileTypeMapper;
+
+	public function __construct(FileTypeMapper $fileTypeMapper)
+	{
+		$this->fileTypeMapper = $fileTypeMapper;
+	}
+
+	public function isAlwaysRead(PropertyReflection $property, string $propertyName): bool
+	{
+		return false;
+	}
+
+	public function isAlwaysWritten(PropertyReflection $property, string $propertyName): bool
+	{
+		return false;
+	}
+
+	public function isInitialized(PropertyReflection $property, string $propertyName): bool
+	{
+		// If the property is public, check for @required on the property itself
+		if (!$property->isPublic()) {
+			return false;
+		}
+
+		if ($property->getDocComment() !== null && $this->isRequiredFromDocComment($property->getDocComment())) {
+			return true;
+		}
+
+		// Check for the attribute version
+		if ($property instanceof PhpPropertyReflection && count($property->getNativeReflection()->getAttributes('Symfony\Contracts\Service\Attribute\Required')) > 0) {
+			return true;
+		}
+
+		return false;
+	}
+
+	public function getAdditionalConstructors(ClassReflection $classReflection): array
+	{
+		$additionalConstructors = [];
+		$nativeReflection = $classReflection->getNativeReflection();
+
+		foreach ($nativeReflection->getMethods() as $method) {
+			if (!$method->isPublic()) {
+				continue;
+			}
+
+			if ($method->getDocComment() !== false && $this->isRequiredFromDocComment($method->getDocComment())) {
+				$additionalConstructors[] = $method->getName();
+			}
+
+			if (count($method->getAttributes('Symfony\Contracts\Service\Attribute\Required')) === 0) {
+				continue;
+			}
+
+			$additionalConstructors[] = $method->getName();
+		}
+
+		return $additionalConstructors;
+	}
+
+	private function isRequiredFromDocComment(string $docComment): bool
+	{
+		$phpDoc = $this->fileTypeMapper->getResolvedPhpDoc(null, null, null, null, $docComment);
+
+		foreach ($phpDoc->getPhpDocNodes() as $node) {
+			// @required tag is available, meaning this property is always initialized
+			if (count($node->getTagsByName('@required')) > 0) {
+				return true;
+			}
+		}
+
+		return false;
+	}
+
+}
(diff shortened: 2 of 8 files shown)


src/Symfony/RequiredAutowiringExtension.php
```diff
@@ -0,0 +1,92 @@
...
+		if (!$property->isPublic()) {
+			return false;
+		}
+
+		if ($property->getDocComment() !== null && $this->isRequiredFromDocComment($property->getDocComment())) {
+			return true;
+		}
+
+		// Check for the attribute version
+		if ($property instanceof PhpPropertyReflection && count($property->getNativeReflection()->getAttributes(Required::class)) > 0) {
```
ondrejmirtes: `Class Symfony\Contracts\Service\Attribute\Required not found.` should be solved by doing `'Symfony\Contracts\Service\Attribute\Required'` as a literal string instead of `::class`.

As for the rest I'm gonna have to think about it and do something in PHPStan itself. Or maybe just ignore them, they're not that severe.

ondrejmirtes (changes requested):

src/Symfony/RequiredAutowiringExtension.php
raalderink: Most issues are solved now, however using the class as a string, will either error out as it being a `string` and not a `class-string`, but when I explicitly type it as `class-string`, it errors with `PHPDoc tag @var with type class-string is not subtype of native type`. It's kind of fighting both sides this way. Is there perhaps a better solution? Stubs maybe?

raalderink: Okay, it took a little bit of effort figuring out the best way to approach this, but I've fixed it now!

raalderink:
@ondrejmirtes could you re-review this PR please? Thank you :)

tests/Symfony/RequiredAutowiringExtensionTest.php
```diff
@@ -0,0 +1,71 @@
...
+final class RequiredAutowiringExtensionTest extends RuleTestCase
+{
+
+	protected function getRule(): Rule
+	{
+		$container = self::getContainer();
+		$container->getServicesByTag(AdditionalConstructorsExtension::EXTENSION_TAG);
+
+		return new UninitializedPropertyRule(
+			new ConstructorsHelper(
```
ondrejmirtes: Please fetch the rule with `$container->getByType(UninitializedPropertyRule::class)`. It should be possible. That way it doesn't have to be in the baseline.

src/Symfony/RequiredAutowiringExtension.php
```diff
@@ -0,0 +1,95 @@
...
+			return true;
+		}
+
+		return false;
+	}
+
+	public function getAdditionalConstructors(ClassReflection $classReflection): array
+	{
+		$additionalConstructors = [];
+		/** @var ReflectionClass $nativeReflection */
```
ondrejmirtes: Please remove the inline `@var`.

src/Symfony/RequiredAutowiringExtension.php
```diff
@@ -0,0 +1,95 @@
...
+		if (!$property->isPublic()) {
+			return false;
+		}
+
+		if ($property->getDocComment() !== null && $this->isRequiredFromDocComment($property->getDocComment())) {
+			return true;
+		}
+
+		// Check for the attribute version
+		if (class_exists(Required::class) && $property instanceof PhpPropertyReflection && count($property->getNativeReflection()->getAttributes(Required::class)) > 0) {
```
ondrejmirtes: The class_exists() check here can be removed.

tests/Type/Symfony/ExtensionTest.php
```diff
@@ -14,49 +14,49 @@ class ExtensionTest extends TypeInferenceTestCase
...
 
-		yield from $this->gatherAssertTypes(__DIR__ . '/data/tree_builder.php');
+		yield from self::gatherAssertTypes(__DIR__ . '/data/tree_builder.php');
 
-		yield from $this->gatherAssertTypes(__DIR__ . '/data/ExampleBaseCommand.php');
-		yield from $this->gatherAssertTypes(__DIR__ . '/data/ExampleOptionCommand.php');
-		yield from $this->gatherAssertTypes(__DIR__ . '/data/ExampleOptionLazyCommand.php');
-		yield from $this->gatherAssertTypes(__DIR__ . '/data/kernel_interface.php');
-		yield from $this->gatherAssertTypes(__DIR__ . '/data/request_get_content.php');
+		yield from self::gatherAssertTypes(__DIR__ . '/data/ExampleBaseCommand.php');
```
ondrejmirtes: These `$this->` to `self::`  are unrelated to the rest of the PR. Please remove them. If you want, you can submit them in a separate PR. Thanks.

ondrejmirtes (changes requested):

tests/Symfony/RequiredAutowiringExtensionTest.php
raalderink: I've done that and removed it from the baseline, however, it still fails on the same thing; 
`Accessing PHPStan\Rules\Properties\UninitializedPropertyRule::class is not covered by backward compatibility promise. The class might change in a minor PHPStan version.`

src/Symfony/RequiredAutowiringExtension.php
raalderink: On PHP 7.3 and lower this will throw the following;
`Call to method getMethods() on an unknown class PHPStan\BetterReflection\Reflection\Adapter\ReflectionEnum.`
I'm not quite sure what to do with this, as I'm confused as to why it would do this only on PHP 7.3 and lower

composer.json
```diff
@@ -15,7 +15,8 @@
   "require": {
     "php": "^7.2 || ^8.0",
     "ext-simplexml": "*",
-    "phpstan/phpstan": "^1.11"
+    "phpstan/phpstan": "^1.11",
+    "symfony/contracts": "^2.2"
```
ondrejmirtes: The extension doesn't really require `symfony/contracts`, please remove this.

raalderink:
I have a couple issues here which I'm having a hard time getting resolved, as they fail on some PHP versions and succeed on others. I've baselined them, but unfortunately `reportUnmatchedIgnoredErrors` is on, which means it will still fail on some and succeed on others.

This following error is actually also baselined in `phpstan-src`, in [enum-adapter-errors.neon](https://github.com/phpstan/phpstan-src/blob/f89852f8c020e7ddc76538f3e6dea510a7271dfd/build/enum-adapter-errors.neon). Due to that, I assume it's a known issue for PHP 7.2 and 7.3, but due to the `reportUnmatchedIgnoredErrors`, I can't really baseline this in this project. How would you want me to solve this issue?
```
-
	message: "#^Call to method getMethods\\(\\) on an unknown class PHPStan\\\\BetterReflection\\\\Reflection\\\\Adapter\\\\ReflectionEnum\\.$#"
	count: 1
	path: src/Symfony/RequiredAutowiringExtension.php
```

Secondly, on the lowest dependency builds, I get the following errors. On the highest dependencies, this builds correctly. What I've done is removed the `class_exists` like you asked, and instead used the `class-string` as string itself. I am having a lot of trouble in finding where this issue lies, and am uncertain of a solution. This is why I had the `class_exists` in the first place, and later tried it by including the actual `symfony-contracts` package that contains it (it makes sense that you don't want that to be included though). Could you please advise on how to approach this and be able to solve this issue?
```
-
	message: "#^Parameter \\#1 \\$name of method PHPStan\\\\BetterReflection\\\\Reflection\\\\Adapter\\\\ReflectionProperty\\:\\:getAttributes\\(\\) expects class\\-string\\|null, string given\\.$#"
	count: 1
	path: src/Symfony/RequiredAutowiringExtension.php

-
	message: "#^Parameter \\#1 \\$name of method PHPStan\\\\BetterReflection\\\\Reflection\\\\Adapter\\\\ReflectionMethod\\:\\:getAttributes\\(\\) expects class\\-string\\|null, string given\\.$#"
	count: 1
	path: src/Symfony/RequiredAutowiringExtension.php
```

These are the only things holding this PR back, so hopefully you can help me out by pointing me towards a solution here. Thanks @ondrejmirtes!

ondrejmirtes (approved):

ondrejmirtes:
Solved it, mainly with https://github.com/phpstan/phpstan-src/commit/2b9af363a3d2a9620c4b9bfb65e81813b69623d0, but also locking contracts package version to a newer one :)

ondrejmirtes:
Thank you.

Issue #89, opened 2020-07-17
ServiceSubscriberInterface, voters and private services issue


vandroids:
# Support question

Hello!
I'm trying to use ServiceSubscriberInterface after reading this article https://symfonycasts.com/screencast/symfony-doctrine/service-subscriber

They suggest to use service locator pattern for the sake of performance in voters.
And the problem is this:

```
namespace App\Voters;

use Psr\Container\ContainerInterface;
use Symfony\Component\Security\Core\Authorization\AccessDecisionManagerInterface;
use Symfony\Component\Security\Core\Authentication\Token\TokenInterface;
use Symfony\Component\Security\Core\Authorization\Voter\Voter;
use Symfony\Contracts\Service\ServiceSubscriberInterface;

class MyVoter extends Voter implements ServiceSubscriberInterface
{
    /**
     * @var ContainerInterface
     */
    private $container;

    public function __construct(ContainerInterface $container)
    {
        $this->container = $container;
    }

    protected function voteOnAttribute($attribute, $subject, TokenInterface $token)
    {
        $this->container->get(AccessDecisionManagerInterface::class)->decide...
    }

    public static function getSubscribedServices()
    {
        return [
            AccessDecisionManagerInterface::class
        ];
    }
}
```
Phpstan tells me that Service "Symfony\Component\Security\Core\Authorization\AccessDecisionManagerInterface" is private.
But my code actually works without any problems.

Can you please help me to find out what to do in this situation?

malarzm:
Just bumped into same issue, seems that https://github.com/phpstan/phpstan-symfony/commit/c340de842c656bdf5ea01eed07fce58ceb6eeb61 wasn't a full fix

specdrum-agc:
@malarzm  there is old deprecated interface `Symfony\Component\DependencyInjection\ServiceSubscriberInterface` in the commit https://github.com/phpstan/phpstan-symfony/commit/c340de842c656bdf5ea01eed07fce58ceb6eeb61 . It has been removed in symfony 4, AFAIR. For now it is `Symfony\Contracts\Service\ServiceSubscriberInterface`

malarzm:
@specdrum-agc thanks for the ping! From what I can see we were originally using the correct interface (god bless GitHub showing links in issues from my private repositories). But also nowadays we no longer have the error in our baseline so I reckon this was fixed somewhen in the past by the way :)

PR #352, opened 2023-05-08
Allow service subscriber to have non-service name keys under certain circumstances.

Basster:
You can define service subscribers with non-service name keys, e.g. 

```php
class SubmissionStrategyFactory implements ServiceSubscriberInterface
{
    public function __construct(private readonly ContainerInterface $locator)
    {
    }

    public function getStrategy(IllustrationSubmission $submission): SubmissionStrategyInterface
    {
        if ($submission->getIllustrationFile()) {
            return $this->locator->get('upload');
        }

        return $this->locator->get('url');
    }

    public static function getSubscribedServices(): iterable
    {
        yield 'upload' => UploadStrategy::class;
        yield 'url' => UrlStrategy::class;
    }
}
```

phpstan currently reports this as `Service "upload" is not registered in the container.`

This PR allows non-service argument names when the argument is a container and its called within a ServiceLocator.

Maybe, this is also related to #89

### DIFF
--- a/src/Rules/Symfony/ContainerInterfaceUnknownServiceRule.php
+++ b/src/Rules/Symfony/ContainerInterfaceUnknownServiceRule.php
@@ -89,8 +89,14 @@ final class ContainerInterfaceUnknownServiceRule implements Rule
 	{
 		$scopeClassReflection = $scope->getClassReflection();
 		return $scopeClassReflection !== null &&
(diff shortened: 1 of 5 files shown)


stof:
As is, this won't solve #89. For that, you would have to do the same detection of ServiceSubscriberInterface in the ContainerInterfacePrivateServiceRule.

PR #353, opened 2023-05-12
Extend Serializer stubs

NicoHaase:
Since https://github.com/symfony/symfony/pull/43982, some interfaces from `symfony/serializer` can use an additional `$context` array. This PR extends the stubs, such that everyone can resolve the deprecation messages in their code without manually declaring the type of `$context`


ondrejmirtes:
Thank you!

Issue #344, opened 2023-04-13
Call to an undefined method Symfony\Component\Validator\ConstraintViolationListInterface::__toString()


MichaelBrauner:
I get this error: 

`Call to an undefined method Symfony\Component\Validator\ConstraintViolationListInterface::__toString()`

But `__toString()` actually should be implemented: https://github.com/symfony/symfony/blob/6.3/src/Symfony/Component/Validator/ConstraintViolationListInterface.php#L24

ondrejmirtes:
The `@method` is missing here, please send a PR that syncs the stub: https://github.com/phpstan/phpstan-symfony/blob/1.2.x/stubs/Symfony/Component/Validator/ConstraintViolationListInterface.stub

ondrejmirtes:
Fixed: https://github.com/phpstan/phpstan-symfony/pull/356

PR #356, opened 2023-06-07
feat: Sync Validor stubs with SF6.1

viviengaetan:
Fix for #344

Adding method __toString() to Validator Stubs according to [real Interface](https://github.com/symfony/symfony/blob/6.1/src/Symfony/Component/Validator/ConstraintViolationListInterface.php#L24)

I added some missing part too but I'm not sure if it can break anything in older Symfony / PHP. I can keep it to only adding method if that's needed

### DIFF
--- a/stubs/Symfony/Component/Validator/ConstraintViolationInterface.stub
+++ b/stubs/Symfony/Component/Validator/ConstraintViolationInterface.stub
@@ -2,6 +2,9 @@
 
 namespace Symfony\Component\Validator;
 
+/**
+ * @method string __toString() Converts the violation into a string for debugging purposes. Not implementing it is deprecated since Symfony 6.1.
+ */
 interface ConstraintViolationInterface
 {
 }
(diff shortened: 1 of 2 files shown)


ondrejmirtes:
Thank you.

VincentLanglet:
HI @ondrejmirtes, 

This (and some other feature maybe) was merge in 1.2.x but doesn't appear in any tags
https://github.com/phpstan/phpstan-symfony/releases

And there are some 1.3.x tags, but the default branch of https://github.com/phpstan/phpstan-symfony seems to be 1.2.x.

Does the 1.2.x branch should be merged into 1.3.x, the 1.3.x used as a default and a new tag done or something else ?

Thanks.

ondrejmirtes:
@VincentLanglet Either I need to do some branch reshuffling, or this will have to wait after PHPStan 1.11 is released this fall.

VincentLanglet:
> @VincentLanglet Either I need to do some branch reshuffling, or this will have to wait after PHPStan 1.11 is released this fall.

Oh, I see https://github.com/phpstan/phpstan-symfony/compare/1.3.2...1.2.x

Your call. I'll ignore my errors until the new release.
Thanks for the quick answer

PR #358, opened 2023-07-08
Fix phpdoc of Twig* callable

VincentLanglet:
This is a Backport of https://github.com/twigphp/Twig/pull/3856

Context can be found here too https://github.com/phpstan/phpstan/discussions/9567

This will solve the issue even if
- symfony does not release a new twig version
- people does not use the latest version (when it's released)


ondrejmirtes:
> symfony does not release a new twig version

I'm pretty sure Symfony will release a new Twig version :)

> People does not use the latest version (when it's released)

They should use the latest version!

Anyway, I don't see a reason for this stub if the fix is already merged over in the original repo. Thanks for understanding.

VincentLanglet:
I thought this was the same than https://github.com/phpstan/phpstan-symfony/pull/318

PR #361, opened 2023-08-28
Add a more precise return type for ParameterBag::keys in stubs

stof:


ondrejmirtes:
Thank you.

stof:
@ondrejmirtes could you do a release including that change ?

ondrejmirtes:
@stof Had to reshuffle the branches a bit but it's released now.

stof:
thanks

PR #368, opened 2023-10-27
Fix stub of Voter::supports() method

### LINKED ISSUE #367: Method MyVoter::supports() has parameter $attribute with no type specified
After update from 1.1.8 to 1.3.4 all custom voters started producing an error, although Symfony's abstract Voter has type annotation for $attribute:

> Method MyVoter::supports() has parameter $attribute with no type specified.

Consider the following simple Voter implementation:

```php
use Symfony\Component\Security\Core\Authorization\Voter\Voter;

class MyVoter extends Voter
{
    protected function supports($attribute, mixed $subject): bool
    {
        return true;
    }

    protected function voteOnAttribute($attribute, mixed $subject, TokenInterface $token): bool
    {
        return true;
    }
}
```

The funny thing is it only gets upset by the `supports` method, but not the `voteOnAttribute` method.

image

luxemate:
closes #367

### DIFF
--- a/stubs/Symfony/Component/Security/Core/Authorization/Voter/Voter.stub
+++ b/stubs/Symfony/Component/Security/Core/Authorization/Voter/Voter.stub
@@ -13,10 +13,11 @@ abstract class Voter implements VoterInterface
     /**
      * Determines if the attribute and subject are supported by this voter.
      *
-     * @param mixed $subject
+     * @phpstan-param TAttribute $attribute
+     * @phpstan-param TSubject $subject
      *
-     * @phpstan-assert-if-true TSubject $subject
      * @phpstan-assert-if-true TAttribute $attribute
+     * @phpstan-assert-if-true TSubject $subject
      *
      * @return bool
      */


stubs/Symfony/Component/Security/Core/Authorization/Voter/Voter.stub
```diff
@@ -13,10 +13,11 @@ abstract class Voter implements VoterInterface
     /**
      * Determines if the attribute and subject are supported by this voter.
      *
-     * @param mixed $subject
+     * @phpstan-param TAttribute $attribute
```
ondrejmirtes, line 16: This is a wrong fix because it's okay to ask about anything in `supports`. The `TAttribute` is typehinted only in `voteOnAttribute` because only valid `TAttribute` can be passed there. But in `supports` you can ask about whatever.

ondrejmirtes (changes requested):

stubs/Symfony/Component/Security/Core/Authorization/Voter/Voter.stub
luxemate: Fixed as per your [comment](https://github.com/phpstan/phpstan-symfony/issues/367#issuecomment-1785362022).

PR #370, opened 2023-10-30
Fix stub of Voter::supports() method

### LINKED ISSUE #367: Method MyVoter::supports() has parameter $attribute with no type specified

luxemate:
closes #367

### DIFF
--- a/stubs/Symfony/Component/Security/Core/Authorization/Voter/Voter.stub
+++ b/stubs/Symfony/Component/Security/Core/Authorization/Voter/Voter.stub
@@ -13,11 +13,11 @@ abstract class Voter implements VoterInterface
     /**
      * Determines if the attribute and subject are supported by this voter.
      *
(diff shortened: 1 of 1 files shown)


ondrejmirtes:
Thank you.

PR #375, opened 2023-12-21
Fix phpstan-src build errors

staabm:
before this PR

```
There was 1 failure:

1) PHPStan\Type\Symfony\ExtensionTest::testFileAsserts with data set "/home/runner/work/phpstan-src/phpstan-src/extension/tests/Type/Symfony/data/cache.php:12" ('type', '/home/runner/work/phpstan-src...he.php', PHPStan\Type\Constant\ConstantStringType Object (...), PHPStan\Type\Constant\ConstantStringType Object (...), 12)
Expected type string, got type '' in /home/runner/work/phpstan-src/phpstan-src/extension/tests/Type/Symfony/data/cache.php on line 12.
Failed asserting that two strings are identical.
--- Expected
+++ Actual
@@ @@
-'string'
+''''
```

### DIFF
--- a/src/Type/Symfony/CacheInterfaceGetDynamicReturnTypeExtension.php
+++ b/src/Type/Symfony/CacheInterfaceGetDynamicReturnTypeExtension.php
@@ -42,6 +42,7 @@ final class CacheInterfaceGetDynamicReturnTypeExtension implements DynamicMethod
 			if ($returnType->isConstantScalarValue()->yes()) {
 				return $returnType->generalize(GeneralizePrecision::lessSpecific());
 			}
+			return $returnType;
 		}
 
 		return null;
--- a/src/Type/Symfony/CacheInterfaceGetDynamicReturnTypeExtension.php
+++ b/src/Type/Symfony/CacheInterfaceGetDynamicReturnTypeExtension.php
@@ -39,10 +39,7 @@ final class CacheInterfaceGetDynamicReturnTypeExtension implements DynamicMethod
 			$returnType = $parametersAcceptor->getReturnType();
 
 			// generalize template parameters
-			if ($returnType->isConstantScalarValue()->yes()) {
-				return $returnType->generalize(GeneralizePrecision::lessSpecific());
-			}
-			return $returnType;
+			return $returnType->generalize(GeneralizePrecision::templateArgument());
 		}
 
 		return null;
--- a/src/Type/Symfony/CacheInterfaceGetDynamicReturnTypeExtension.php
+++ b/src/Type/Symfony/CacheInterfaceGetDynamicReturnTypeExtension.php
@@ -1,0 +1,50 @@
+<?php declare(strict_types = 1);
+
+namespace PHPStan\Type\Symfony;
+
+use PhpParser\Node\Expr\MethodCall;
+use PHPStan\Analyser\Scope;
+use PHPStan\Reflection\MethodReflection;
+use PHPStan\Reflection\ParametersAcceptorSelector;
+use PHPStan\Type\DynamicMethodReturnTypeExtension;
+use PHPStan\Type\GeneralizePrecision;
+use PHPStan\Type\Type;
+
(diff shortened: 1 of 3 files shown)


ondrejmirtes:
This is not the right fix. The failing test actually shows a real issue. PHPStan dynamic return type extension needs to be written instead.

src/Type/Symfony/CacheInterfaceGetDynamicReturnTypeExtension.php
```diff
@@ -0,0 +1,45 @@
...
+	}
+
+	public function getTypeFromMethodCall(MethodReflection $methodReflection, MethodCall $methodCall, Scope $scope): ?Type
+	{
+		if (!isset($methodCall->getArgs()[1])) {
+			return null;
+		}
+
+		$callbackReturnType = $scope->getType($methodCall->getArgs()[1]->value);
+		if ($callbackReturnType instanceof ClosureType) {
```
ondrejmirtes: CacheInterface accepts any callable, `instanceof *Type` is always wrong.

`Type` has `getCallableParametersAcceptors`, check `isCallable()->yes()` and then call this method.

ondrejmirtes (changes requested):
And please rebase this on top of 1.3.x.

composer.json
```diff
@@ -15,7 +15,7 @@
   "require": {
     "php": "^7.2 || ^8.0",
     "ext-simplexml": "*",
-    "phpstan/phpstan": "^1.10.36"
+    "phpstan/phpstan": "^1.10.49"
```
staabm: since the behaviour we fix with this PR only affects 1.10.49+
refs https://github.com/phpstan/phpstan-src/pull/2818

ondrejmirtes: I disagree with this change. The dynamic return type extension is valid and works for older PHPStan versions too.

ondrejmirtes (changes requested):

src/Type/Symfony/CacheInterfaceGetDynamicReturnTypeExtension.php
```diff
@@ -0,0 +1,50 @@
...
+				$scope,
+				$methodCall->getArgs(),
+				$callbackReturnType->getCallableParametersAcceptors($scope)
+			);
+			$returnType = $parametersAcceptor->getReturnType();
+
+			// generalize template parameters
+			if ($returnType->isConstantScalarValue()->yes()) {
+				return $returnType->generalize(GeneralizePrecision::lessSpecific());
+			}
```
ondrejmirtes: Please test a scenario where this is not a scalar. The `$returnType` should still be used.

ondrejmirtes: I'd rather `return $returnType` here too.

ondrejmirtes (changes requested):

tests/Type/Symfony/data/cache.php
```diff
@@ -12,6 +12,15 @@ function testCacheCallable(\Symfony\Contracts\Cache\CacheInterface  $cache): voi
 	assertType('string', $result);
 };
 
+/**
+ * @param callable():string $fn
+ */
+function testNonScalarCacheCallable(\Symfony\Contracts\Cache\CacheInterface $cache, callable $fn): void {
+	$result = $cache->get('foo', $fn);
+
+	assertType('string', $result);
```
ondrejmirtes, line 21: I feel like `non-empty-string` should be generalized to `string`. Which means we'll likely have to always generalize by `GeneralizePrecision::templateArgument()`.

ondrejmirtes (changes requested):

tests/Type/Symfony/data/cache.php
staabm: wasn't aware `GeneralizePrecision::templateArgument()` is a thing 😅

ondrejmirtes (approved):

ondrejmirtes:
Thank you.

PR #384, opened 2024-02-29
Narrow type for Extension::getConfiguration if class exists

RobertMe:
The base extension class automatically creates a Configuration instance when a Configuration class exists in the namespace of the extension. But PHPStan obviously doesn't understand this behaviour and always assumes that `getConfiguration()` returns `ConfigurationInterface|null` meaning that the default pattern to get and parse the configuration reports an error.

I.e.:
```php
namespace Foo;

class SomeExtension extends Extension
{
    public function load(array $configs, ContainerBuilder $container): void
    {
        $configuration = $this->getConfiguration($configs, $container);
        $config = $this->processConfiguration($configuration, $configs);
    }
}
```
results in an error because `processConfiguration()` doesn't accept `ConfigurationInterface|null`. But when a `Configuration` class exists in the same namespace as the `Extension` class (so `Foo\Extension`) an instance of it is returned.

This `DynamicReturnTypeExtension` overrides the return type of `Extension::getConfiguration()` so it automatically narrows the return type in case `getConfiguration()` is not overriden and a `Configuration` class exists. So that in the given example `getConfiguration()` doesn't return `ConfigurationInterface|null` anymore but `Foo\Configuration` and there is no error on calling `processConfiguration()`.

### DIFF
--- a/src/Type/Symfony/ExtensionGetConfigurationReturnTypeExtension.php
+++ b/src/Type/Symfony/ExtensionGetConfigurationReturnTypeExtension.php
@@ -1,0 +1,105 @@
+<?php declare(strict_types = 1);
+
+namespace PHPStan\Type\Symfony;
+
+use PhpParser\Node\Expr\MethodCall;
+use PHPStan\Analyser\Scope;
+use PHPStan\Reflection\ClassReflection;
+use PHPStan\Reflection\MethodReflection;
+use PHPStan\Reflection\ReflectionProvider;
+use PHPStan\Type\DynamicMethodReturnTypeExtension;
+use PHPStan\Type\NullType;
+use PHPStan\Type\ObjectType;
+use PHPStan\Type\Type;
+use PHPStan\Type\TypeCombinator;
+use function str_contains;
+use function strrpos;
+use function substr_replace;
+
+class ExtensionGetConfigurationReturnTypeExtension implements DynamicMethodReturnTypeExtension
+{
+
+	/** @var ReflectionProvider */
+	private $reflectionProvider;
+
+	public function __construct(ReflectionProvider $reflectionProvider)
+	{
+		$this->reflectionProvider = $reflectionProvider;
+	}
+
+	public function getClass(): string
+	{
+		return 'Symfony\Component\DependencyInjection\Extension\Extension';
+	}
+
+	public function isMethodSupported(MethodReflection $methodReflection): bool
+	{
+		return $methodReflection->getName() === 'getConfiguration'
+			&& $methodReflection->getDeclaringClass()->getName() === 'Symfony\Component\DependencyInjection\Extension\Extension';
+	}
+
+	public function getTypeFromMethodCall(
+		MethodReflection $methodReflection,
+		MethodCall $methodCall,
+		Scope $scope
+	): ?Type
+	{
+		$types = [];
+		$extensionType = $scope->getType($methodCall->var);
+		$classes = $extensionType->getObjectClassNames();
+
+		foreach ($classes as $extensionName) {
+			if (str_contains($extensionName, "\0")) {
+				$types[] = new NullType();
+				continue;
+			}
+
+			$lastBackslash = strrpos($extensionName, '\\');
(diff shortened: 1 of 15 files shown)


tests/Type/Symfony/data/extension/with-configuration/WithConfigurationExtension.php
```diff
@@ -0,0 +1,17 @@
...
+
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use \Symfony\Component\DependencyInjection\Extension\Extension;
+
+class WithConfigurationExtension extends Extension
+{
+	public function load(array $configs, ContainerBuilder $container): void
+	{
+		\PHPStan\Testing\assertType(
+			'PHPStan\Type\Symfony\Extension\WithConfiguration\Configuration',
```
ondrejmirtes: You could do `Configuration::class` here.

src/Type/Symfony/ExtensionGetConfigurationReturnTypeExtension.php
```diff
@@ -0,0 +1,72 @@
...
+		if ($lastBackslash === false) {
+			return new NullType();
+		}
+
+		$configurationName = substr_replace($extensionName, '\Configuration', $lastBackslash);
+		if (!$this->reflectionProvider->hasClass($configurationName)) {
+			return new NullType();
+		}
+
+		// TODO: return NullType if configuration class has constructor with required arguments
```
ondrejmirtes: Please solve this todo, or remove it.

src/Type/Symfony/ExtensionGetConfigurationReturnTypeExtension.php
```diff
@@ -0,0 +1,72 @@
...
+	}
+
+	public function getTypeFromMethodCall(
+		MethodReflection $methodReflection,
+		MethodCall $methodCall,
+		Scope $scope
+	): ?Type
+	{
+		$extensionType = $scope->getType($methodCall->var);
+		$classes = $extensionType->getObjectClassNames();
```
ondrejmirtes, line 49: We can solve this for union types too. Don't return early if it's not 1, iterate over the class names and create a union.

ondrejmirtes (changes requested):

tests/Type/Symfony/data/extension/with-configuration/WithConfigurationExtension.php
RobertMe: Will do. I think I based it on some other test fixture, which also used a string. But if `::class` works and is preferred I will obviously change it 👍🏻

src/Type/Symfony/ExtensionGetConfigurationReturnTypeExtension.php
RobertMe: If you can provide some pointers on how to determine this I will try to implement it :) I haven't been able to find some example on how to determine / inspect such information.

ondrejmirtes: You can get ClassReflection from ReflectionProvider and ask about constructor from there.

src/Type/Symfony/ExtensionGetConfigurationReturnTypeExtension.php
RobertMe: True. But I don't think it's common. I believe that `->getConfiguration()` is only called on `$this` (/by the extension itself), or by the Symfony framework. So my reasoning was that handling it for other cases (/unions) makes the code more complex while not being of any benefit.

But I'll update the code so it handles that as well 👍🏻

src/Type/Symfony/ExtensionGetConfigurationReturnTypeExtension.php
RobertMe: That was the easy part which I did find out already :) Anyway, I'm now iterating over the variants => parameters => checking `isOptional` on every one of them. Not sure if there is an easier way to do this though.

ondrejmirtes:
Thank you!

PR #386, opened 2024-03-13
Add param out for propertyAccessor

VincentLanglet:
When setting value to an object the object type does not change.

When setting value to an object, the array values change. But at least we know it's an array and not `array|object`.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -15,7 +15,7 @@
   "require": {
     "php": "^7.2 || ^8.0",
     "ext-simplexml": "*",
-    "phpstan/phpstan": "^1.10.36"
+    "phpstan/phpstan": "^1.10.62"
   },
   "conflict": {
     "symfony/framework-bundle": "<3.0"
--- a/stubs/Symfony/Component/PropertyAccess/PropertyAccessorInterface.stub
+++ b/stubs/Symfony/Component/PropertyAccess/PropertyAccessorInterface.stub
@@ -1,0 +1,21 @@
+<?php
+
+namespace Symfony\Component\PropertyAccess;
+
+interface PropertyAccessorInterface
+{
+
+    /**
+     * @phpstan-template T of object|array<mixed>
+     * @phpstan-param T &$objectOrArray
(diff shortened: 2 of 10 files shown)


composer.json
```diff
@@ -15,7 +15,7 @@
   "require": {
     "php": "^7.2 || ^8.0",
     "ext-simplexml": "*",
-    "phpstan/phpstan": "^1.10.36"
+    "phpstan/phpstan": "^1.10.62"
```
VincentLanglet, line 18: For conditional phpstan-param-out

ondrejmirtes:
Thank you.

stubs/Symfony/Component/PropertyAccess/PropertyAccessorInterface.stub
```diff
@@ -0,0 +1,23 @@
...
+
+namespace Symfony\Component\PropertyAccess;
+
+interface PropertyAccessorInterface
+{
+
+    /**
+     * @phpstan-template T of object|array<mixed>
+     * @phpstan-param T &$objectOrArray
+     * @phpstan-param-out ($objectOrArray is object ? T : array<mixed>) $objectOrArray
```
herndlm, line 11: hi, just for my understanding - this could be simply `@phpstan-param-out T $objectOrArray` as well, right?

I had a local stub with something like that and got an error when updating because of this PR -> could remove my local stub -> thx! :)

VincentLanglet:
```
@phpstan-template T of object|array<mixed>
@phpstan-param T &$objectOrArray
@phpstan-param-out T $objectOrArray
```
doesn't give the same beahvior

It fails 
```
$array = [1 => 'ea'];
$propertyAccessor->setValue($array, 'foo', 'bar');
assertType('array', $array);
```
You'll get `array{1: 'ea'}` instead, which is wrong.

herndlm: Ah OK, makes sense. Thx!

PR #388, opened 2024-04-11
Add stub for Compound::getConstraints()

derrabus:
This PR fixes a error people get when implementing compound constraints:

> Method App\Validator\Constraints\MyConstraint::getConstraints() has parameter $options with no value type specified in iterable type array.

Symfony does not document the structure of the `$options` array, which can be an array of anything.


ondrejmirtes:
Thank you! :)

VincentLanglet:
> Symfony does not document the structure of the `$options` array, which can be an array of anything.

Looking at the usage, I think it should be `array<string, mixed>` rather than `array<mixed>`

I created https://github.com/symfony/symfony/pull/54574 to validate the point @derrabus.

derrabus:
Thank you for opening the PR upstream. I've left a comment.

PR #389, opened 2024-04-13
Synchronize Compound::getConstraints() with Symfony phpdoc

VincentLanglet:
Pr https://github.com/symfony/symfony/pull/54574 was accepted.

Which means the phpdoc should be `array<string, mixed>` and not `array<mixed>`
https://github.com/symfony/symfony/pull/54574/files#diff-bb41b84020d311b1a23f6a18724be831e6d9789dd44dc29ed024cbe13ae1a6b1R49


ondrejmirtes:
Thank you.

PR #390, opened 2024-04-29
make creation of ServiceMap and ParameterMap lazy

RobertMe:
Currently one can not use PHPStans `bootstrapFiles` to create the container (dump) which could then be ready by
`symfony.containerXmlPath`. This as the order of initialization results in the `XmlServiceMapFactory` / `XmlParameterMapFactory` `create` methods being called before the `bootstrapFiles` are executed. By making the `ServiceMap` / `ParameterMap` lazy these will only be created when actually used, which is after the `bootstrapFiles` have been executed.

### DIFF
--- a/src/Symfony/ParameterMap.php
+++ b/src/Symfony/ParameterMap.php
@@ -18,6 +18,6 @@ interface ParameterMap
 	/**
 	 * @return array<string>
 	 */
-	public static function getParameterKeysFromNode(Expr $node, Scope $scope): array;
+	public function getParameterKeysFromNode(Expr $node, Scope $scope): array;
 
 }
--- a/src/Symfony/ServiceMap.php
+++ b/src/Symfony/ServiceMap.php
@@ -15,6 +15,6 @@ interface ServiceMap
 
 	public function getService(string $id): ?ServiceDefinition;
 
-	public static function getServiceIdFromNode(Expr $node, Scope $scope): ?string;
+	public function getServiceIdFromNode(Expr $node, Scope $scope): ?string;
 
 }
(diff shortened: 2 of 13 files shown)


src/Symfony/ParameterMap.php
```diff
@@ -18,6 +18,6 @@ public function getParameter(string $key): ?ParameterDefinition;
 	/**
 	 * @return array<string>
 	 */
-	public static function getParameterKeysFromNode(Expr $node, Scope $scope): array;
+	public function getParameterKeysFromNode(Expr $node, Scope $scope): array;
```
RobertMe, line 21: This might be considered a breaking change and thus require a major version bump?

src/Symfony/ServiceMap.php
```diff
@@ -15,6 +15,6 @@ public function getServices(): array;
 
 	public function getService(string $id): ?ServiceDefinition;
 
-	public static function getServiceIdFromNode(Expr $node, Scope $scope): ?string;
+	public function getServiceIdFromNode(Expr $node, Scope $scope): ?string;
```
RobertMe, line 18: Same as above for `ParameterMap` change, this might be considered a breaking change and thus require a major version bump?

RobertMe:
Whoops, changes are PHP 7.2 and 7.3 incompatible. I will fix them if considered mergable etc. But _if_ the changes to the interface are decided to be a breaking change and require a major version bump it _might_ be considered to bump the minimum versions as well?

ondrejmirtes:
> Currently one can not use PHPStans bootstrapFiles to create the container (dump) which could then be ready

this is okay. I’m not willing to merge anything that would make the code more complex, because you can generate the container dump before running PHPStan. It’s been done like that for many years and people don’t have a problem with that.

For example you can use make or Composer scripts to still invoke a single command.

Issue #256, opened 2022-02-23
Required arguments and option can be null in interact


alexander-schranz:
In the `interact` method of the symfony console command its possible to ask for required attributes. In that context a required argument can still be null.

The following code will tell that the null check does not work:

e.g.:

```php
    protected function configure(): void
    {
        $this->addArgument('market', InputArgument::REQUIRED);
    }
    
    protected function interact(InputInterface $input, OutputInterface $output): void
    {
        $market = $input->getArgument('market'); // here market can still be `null|string`

        if ($market === null) {
            $input->setArgument('market', $ui->ask('Which market you want to use for the import?'));
        }
    }
    protected function execute(InputInterface $input, OutputInterface $output): void
    {
        $market = $input->getArgument('market'); // here market can only be a string
    }
```

When using `!$market` rector will actually refractor it to `$test === '' && $test === '0'` so that is also not correct.

If possible it would be great that depending if the argument is used inside interact it should be nullable else it should be the same as currently. Not sure if that is possible via phpstan.

ondrejmirtes:
Sorry, I worry this is out of scope for PHPStan to understand such code.

ondrejmirtes:
You can do this in your `execute` method:

```php
/** @var string $market */
$market = $input->getArgument('market');
```

PR #391, opened 2024-05-20
Update getArgument return type in interact method

VincentLanglet:
Solve https://github.com/phpstan/phpstan-symfony/issues/256 (cc @alexander-schranz)

By design, all the arguments can be null in Command::interact method.
Doing
```
/** @var string|null $market */
$market = $input->getArgument('market');
```
in the interact method is reported as an error by PHPStan so there is currently no satisfying solution except ignoring the error. And updating the dynamic return type extension seems doable.

### DIFF
--- a/src/Type/Symfony/InputInterfaceGetArgumentDynamicReturnTypeExtension.php
+++ b/src/Type/Symfony/InputInterfaceGetArgumentDynamicReturnTypeExtension.php
@@ -10,6 +10,7 @@ use PHPStan\Symfony\ConsoleApplicationResolver;
 use PHPStan\Type\ArrayType;
 use PHPStan\Type\DynamicMethodReturnTypeExtension;
 use PHPStan\Type\IntegerType;
+use PHPStan\Type\NullType;
 use PHPStan\Type\StringType;
 use PHPStan\Type\Type;
 use PHPStan\Type\TypeCombinator;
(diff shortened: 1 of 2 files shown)


Griffon-Weglot (approved):

src/Type/Symfony/InputInterfaceGetArgumentDynamicReturnTypeExtension.php
```diff
@@ -76,6 +77,11 @@ public function getTypeFromMethodCall(MethodReflection $methodReflection, Method
 			}
 		}
 
+		$method = $scope->getFunction();
+		if ($method instanceof MethodReflection && $method->getName() === 'interact') {
```
ondrejmirtes: You should also check the declaring class of the method. You might not be in a console command at all.

ondrejmirtes (changes requested):

src/Type/Symfony/InputInterfaceGetArgumentDynamicReturnTypeExtension.php
VincentLanglet: I added a new check https://github.com/phpstan/phpstan-symfony/pull/391/commits/1dfe11998bcee3529ffc18332a00c385f4e14bb0

I'm not sure if there is a more optimised method for this.

ondrejmirtes:
Thank you.

PR #392, opened 2024-05-23
Add DefinitionConfigurator stub for rootNode ArrayNodeDefinition return type

### LINKED ISSUE #278: [6.1] [AbstractBundle] Call to an undefined method `NodeDefinition::children()`
When using the new Symfony 6.1 `AbstractBundle` introduced in https://github.com/symfony/symfony/pull/43701 I'm getting the following error:

```php
final class GraphQLBundle extends AbstractBundle
{
    public function configure(DefinitionConfigurator $definition) : void
    {
        $definition
            ->rootNode()
            ->children()  // Call to an undefined method Symfony\Component\Config\Definition\Builder\NodeDefinition::children().
                ->integerNode('debug_flag')
                    ->defaultValue(DebugFlag::NONE)
                    ->isRequired()
                    ->end()
            ->end();
    }
}
```

I think this is because [`DefinitionConfigurator::rootNode()` returns `NodeDefinition|ArrayNodeDefinition`](https://github.com/symfony/symfony/blob/a10071bd657c350bb8f995361643072a97ff5819/src/Symfony/Component/Config/Definition/Configurator/DefinitionConfigurator.php#L38-L41). It will always return a `ArrayNodeDefinition`. Maybe the return type should be changed in Symfony to reflect that?

The [TreeBuilderGetRootNodeDynamicReturnTypeExtension](https://github.com/phpstan/phpstan-symfony/blob/1.2.x/src/Type/Symfony/Config/TreeBuilderGetRootNodeDynamicReturnTypeExtension.php) won't work anymore because the TreeBuilder instance is created by Symfony for you.

I don't really know what's the best way to solve this in PHPStan. Create a new extension that overwrites the `rootNode()` method's return type?

/cc @yceruto

andyexeter:
Fixes #278 

The DefinitionConfigurator::rootNode method always returns an ArrayNodeDefinition, but without this PR when we use this code:

```php
use Symfony\Component\Config\Definition\Configurator\DefinitionConfigurator;

class ExampleBundle extends AbstractBundle
{
    public function configure(DefinitionConfigurator $definition): void
    {
        $definition->rootNode()
            ->children()
            ->end()
        ;
    }
```

We get the following error:

```
Call to an undefined method Symfony\Component\Config\Definition\Builder\NodeDefinition::children().
```

### DIFF
--- a/extension.neon
+++ b/extension.neon
@@ -32,6 +32,7 @@ parameters:
 		- stubs/Symfony/Bundle/FrameworkBundle/Test/TestContainer.stub
 		- stubs/Symfony/Bundle/SecurityBundle/DependencyInjection/Security/Factory/AuthenticatorFactoryInterface.stub
 		- stubs/Symfony/Bundle/SecurityBundle/DependencyInjection/Security/Factory/FirewallListenerFactoryInterface.stub
+		- stubs/Symfony/Component/Config/Definition/Configurator/DefinitionConfigurator.stub
 		- stubs/Symfony/Component/Console/Command.stub
 		- stubs/Symfony/Component/Console/Exception/ExceptionInterface.stub
 		- stubs/Symfony/Component/Console/Exception/InvalidArgumentException.stub
--- a/extension.neon
+++ b/extension.neon
@@ -32,6 +32,7 @@ parameters:
 		- stubs/Symfony/Bundle/FrameworkBundle/Test/TestContainer.stub
 		- stubs/Symfony/Bundle/SecurityBundle/DependencyInjection/Security/Factory/AuthenticatorFactoryInterface.stub
 		- stubs/Symfony/Bundle/SecurityBundle/DependencyInjection/Security/Factory/FirewallListenerFactoryInterface.stub
+		- stubs/Symfony/Component/Config/Definition/Builder/ArrayNodeDefinition.stub
 		- stubs/Symfony/Component/Config/Definition/Configurator/DefinitionConfigurator.stub
 		- stubs/Symfony/Component/Console/Command.stub
 		- stubs/Symfony/Component/Console/Exception/ExceptionInterface.stub
--- a/stubs/Symfony/Component/Config/Definition/Builder/ArrayNodeDefinition.stub
+++ b/stubs/Symfony/Component/Config/Definition/Builder/ArrayNodeDefinition.stub
@@ -1,0 +1,8 @@
+<?php
+
+namespace Symfony\Component\Config\Definition\Builder;
+
+class ArrayNodeDefinition
+{
+
+}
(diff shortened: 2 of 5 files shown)


andyexeter:
I'm not sure why the build is failing

Seems to be an issue with result cache. Running `make phpstan` with an empty result cache locally causes the same error, but running it with a filled result cache returns no errors.

andyexeter:
Ah, I had to add the ArrayNodeDefinition stub. CI is passing now :)

ondrejmirtes:
According to comment in code (https://github.com/symfony/symfony/blob/1f90bc30c32b375ed2cdc3234f56d26b8b99cb53/src/Symfony/Component/Config/Definition/Builder/TreeBuilder.php#L33) this isn't always ArrayNodeDefinition. Can you elaborate on that?

andyexeter:
Sure. `ArrayNodeDefinition` is always returned by `DefinitionConfigurator::rootNode` because the `DefinitionConfigurator` `TreeBuilder` instance is constructed with the default `array` type: 

https://github.com/symfony/symfony/blob/1f90bc30c32b375ed2cdc3234f56d26b8b99cb53/src/Symfony/Component/Config/Definition/Configuration.php#L36

https://github.com/symfony/symfony/blob/1f90bc30c32b375ed2cdc3234f56d26b8b99cb53/src/Symfony/Component/Config/Definition/Builder/TreeBuilder.php#L26

ondrejmirtes:
But if someone passes a different `$type` than array this assumption does not hold.

andyexeter:
The TreeBuilder is constructed within Symfony’s config component so the `$type` value cannot be changed. 

The rootNode method of DefinitionConfigurator will always return an ArrayNodeDefinition because that’s how its TreeBuilder is constructed. End developers cannot change the construction of the TreeBuilder

ondrejmirtes:
1) DefinitionConfigurator accepts TreeBuilder.
2) `DefinitionConfigurator::rootNode()` calls `$this->treeBuilder->getRootNode()`.
3) `TreeBuilder::getRootNode()` returns `$this->root`.
4) `$this->root` is assigned like this: `$this->root = $builder->node($name, $type)->setParent($this);`
5) Therefore the type depends on how `TreeBuilder` is constructed. `$type` is passed in the constructor and can be different from `array`.

andyexeter:
But the`DefinitionConfigurator` is only ever constructed in `Configuration::getTreeBuilder`, which doesn't pass a type.

I see your point though, it _could_ be constructed elsewhere. Would a dynamic return type extension work here? Similar to [TreeBuilderGetRootNodeDynamicReturnTypeExtension](https://github.com/phpstan/phpstan-symfony/blob/51183fefbaf4713aa81eddbd273dc59dd5e5ff71/src/Type/Symfony/Config/TreeBuilderGetRootNodeDynamicReturnTypeExtension.php)?

ondrejmirtes:
Hey, if you're really confident the stub should look like this, it should be easy to convince Symfony to update their return type.

But I don't think your assumption is correc,t and that's why I'm closing this PR. In my opinion the current behaviour is correct and you shouldn't rely on ArrayNodeDefinition being always returned by the `DefinitionConfigurator::rootNode()` method.

PR #394, opened 2024-05-29
Support for alias to inlined service

b-viguier:
# Why
When [stacking decorators](https://symfony.com/doc/current/service_container/service_decoration.html#stacking-decorators), all inlined services are prefixed by a `.`. This character is explicitly removed in the service map since [this commit](https://github.com/phpstan/phpstan-symfony/commit/74da071c10482f509dcb443e26fb52a4e9511556).

But then, when browsing aliases, the Xml factory is not able to find the target service in the map, because the alias definition contains the `.`, but not the stored service. The we got following false-positive 
```
Service "<...>" is not registered in the container.
```

# How
I added the same log for `alias` field that the one for `id`, and it fixed the issue locally. I added a test for this use case.

⚠️ I'm not sure why this `.` has been removed in first place, maybe there are other (untested?) side-effects... at least the discussion is opened 🙂 

Thanks a lot for your work 🙇

### DIFF
--- a/src/Symfony/XmlServiceMapFactory.php
+++ b/src/Symfony/XmlServiceMapFactory.php
@@ -48,11 +48,11 @@ final class XmlServiceMapFactory implements ServiceMapFactory
 			}
 
 			$service = new Service(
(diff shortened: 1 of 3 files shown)


ondrejmirtes:
Thank you!

PR #398, opened 2024-06-19
Improve Command::initialize and `command` argument inferences

VincentLanglet:
This follow the same idea of https://github.com/phpstan/phpstan-symfony/pull/391.

When looking at the execute method of the Symfony command
https://github.com/symfony/symfony/blob/c0e30bb06bb4c12efc8bad13f7cb732954cb0182/src/Symfony/Component/Console/Command/Command.php#L231-L280

The order of the call are
```
$this->initialize($input, $output);
$this->interact($input, $output);
if ($input->hasArgument('command') && null === $input->getArgument('command')) {
     $input->setArgument('command', $this->getName());
}
$input->validate();
$this->execute($input, $output);
```

Which means that:
- In the initialize method the `command` argument might be missing
- In the interact method the `command` argument might be missing
- In the initialize method, like in the interact one` option might be missing since they are not yet validated.


ondrejmirtes:
Thank you.

PR #402, opened 2024-07-18
Fix internal error when DIC param map has > 256 items

janedbal:
The failure happened [here](https://github.com/phpstan/phpstan-src/blob/5277630/src/Type/Constant/ConstantArrayType.php#L212).

-----

Generally, this feels like a trap inside PHPStan. If any `ConstantArrayType` will fail when used in `implode` or just with any usage of `TypeUtils::flattenTypes`, it is just not safe to create it by constructor and the degradation logic in `ConstantArrayTypeBuilder` should be imo moved to `ConstantArrayType` constructor. 

Or maybe just `getAllArrays` needs to be fixed? IDK where is the proper place to fix this.


ondrejmirtes:
I feel like we should make a rule in PHPStan so that ConstantArrayTypeBuilder is used instead of `new ConstantArrayType` when the size of the created array is unknown. Because:

> should be imo moved to ConstantArrayType constructor.

That's too late, you can't return a different type from a class constructor.

Otherwise I think this is the right fix. Thank you.

PR #403, opened 2024-08-12
Add stub for AbstractController::createForm()

zacharylund:


stubs/Symfony/Bundle/FrameworkBundle/Controller/AbstractController.stub
```diff
@@ -0,0 +1,24 @@
...
+
+namespace Symfony\Bundle\FrameworkBundle\Controller;
+
+use Symfony\Component\Form\FormInterface;
+use Symfony\Component\Form\FormTypeInterface;
+use Symfony\Contracts\Service\ServiceSubscriberInterface;
+
+abstract class AbstractController implements ServiceSubscriberInterface
+{
+	/**
```
ondrejmirtes, line 11: This is a non-trivial stub. Can you please add some tests (by adding a new file) to https://github.com/phpstan/phpstan-symfony/blob/ee88a01bc48f608143d3376802ec952270737cb8/tests/Type/Symfony/ExtensionTest.php#L5 and use assertType to try out different branches of the conditional return type? Thanks.

ondrejmirtes (changes requested):

stubs/Symfony/Bundle/FrameworkBundle/Controller/AbstractController.stub
zacharylund: I modeled this stub after FormFactoryInterface since that's what AbstractController calls. It seemed appropriate to add this test to that file.

ondrejmirtes (approved):

ondrejmirtes:
Thank you!

Issue #207, opened 2021-11-11
Support for Symfony HandleTrait


alexander-schranz:
I think it would be great if the extension could automatically recognize the type of [`HandleTrait`](https://github.com/symfony/symfony/blob/5.4/src/Symfony/Component/Messenger/HandleTrait.php) of symfony.

The given message to the handle trait example:

```php
$this->handle(new CreateUserMessage());
```

does only have 1 Handler which return type the return type of the handle trait will be. Example:

```php
class CreateUserMessageHandler {
    public function __invoke(CreateUserMessage $message): User
    {
        return new User($message->getUserName());
    }
}
```

The `bin/console debug:messenger` already shows the mapping between Message and Handlers.

The `HandleStamp` contains under the `getResult` method the result of every handler: https://github.com/symfony/symfony/blob/8eed84bb2df613101949ca2e786ac383d2f349fd/src/Symfony/Component/Messenger/HandleTrait.php#L45. The HandleTrait only allows one handler per message, that make the result recognize able. Not sure if its even possible to overwrite the result of a trait. Sadly my skillset for providing this myself is to low, but would think it would be a great feature.

Currenty workaround is:

```php
/** @var User */
$user = $this->handle(new CreateUserMessage());
```

The first parameter of `console.command.messenger_debug` service would contain the mapping between message and handler so we can guess the type from there.

bnowak:
Did anybody take a look on that so far to give some insights? Today, I'm going to start working on this.

bnowak:
So far, I researched that in current version not having of [this](https://github.com/phpstan/phpstan/issues/5761) feature request, it could be a blocker.

but I'm keep digging :wink:

bnowak:
I'm happy to announce that PR for that is done and ready for review [here](https://github.com/phpstan/phpstan-symfony/pull/404) :)

Feel free to review and give your feedback, thank you :)

bnowak:
There's also another QueryBus-classes related topic which I wanted to cover with reusing this feature. It's described [in this comment](https://github.com/phpstan/phpstan-symfony/pull/404#issuecomment-2419761831). If you have idea how to solve it, please share it here :) Thanks, again

ondrejmirtes:
Implemented https://github.com/phpstan/phpstan-symfony/pull/404

PR #404, opened 2024-08-13
Support for Messenger HandleTrait return types

bnowak:
The goal is to cover https://github.com/phpstan/phpstan-symfony/issues/207.

That's my very first PR in phpstan-symfony plugin, so any feedback/tips/advices are very welcome :)


ondrejmirtes (changes requested):
Instead of DynamicMethodReturnTypeExtension you can use https://apiref.phpstan.org/1.11.x/PHPStan.Type.ExpressionTypeResolverExtension.html which is like that but lets you override any expression type.

bnowak:
@ondrejmirtes Thanks for the tip about `ExpressionTypeResolverExtension`! It resolved my blocker issue about not supported traits in `DynamicMethodReturnTypeExtension`.

It took me some time, however finally I've got ready to review version of this PR :) Please take a look in you free time :)

Thank you very much!

bnowak:
As static analyzing should works now fine with the result of `HandleTrait::handle`, but only in classes which have statically typed query passed - like in this SF-adjusted example:
```php
// src/Action/ListItems.php
namespace App\Action;

use App\Message\ListItemsQuery;
use App\MessageHandler\ListItemsQueryResult;
use Symfony\Component\Messenger\HandleTrait;
use Symfony\Component\Messenger\MessageBusInterface;

class ListItems
{
    use HandleTrait;

    public function __construct(
        private MessageBusInterface $messageBus,
    ) {
    }

    public function __invoke(): void
    {
        $result = $this->handle(new ListItemsQuery(/* ... */));
        // $result should be automatically recognized as `ListItemsQueryResult` now by phpstan via SF configuration

        // Do something with the result
        // ...
    }
}
```

However, my target goal was to cover QueryBus classes with dynamic return types based on dynamic query, something like:
```php
// src/MessageBus/QueryBus.php
namespace App\MessageBus;

use Symfony\Component\Messenger\Envelope;
use Symfony\Component\Messenger\HandleTrait;
use Symfony\Component\Messenger\MessageBusInterface;

class QueryBus
{
    use HandleTrait;

    public function __construct(
        private MessageBusInterface $messageBus,
    ) {
    }

    // wherever that method is called, phpstan should recognize return type based on passed query
    public function query($query)
    {
        return $this->handle($query);
    }
}
```

Maybe it's trivial and doable with already existing php-doc annotation or more complex to write some next extension for that, but I don't know how to connect the dots and reuse covered typing of `HandleTrait::handle` which is already in place. I'm happy to address that in separate PR (if will be needed to achieve that) :)
Any thoughts/ideas are more than welcome :)
/cc @ondrejmirtes

src/Symfony/MessageMap.php
```diff
@@ -0,0 +1,26 @@
+<?php declare(strict_types = 1);
+
+namespace PHPStan\Symfony;
+
+use PHPStan\Type\Type;
+
+final class MessageMap
+{
+
+	/** @var array<string, Type> */
```
KrzysztofCzerwinski, line 10: suggestion: `class-string`

src/Symfony/MessageMap.php
```diff
@@ -0,0 +1,26 @@
...
+	private $messageMap;
+
+	/**
+	 * @param array<string, Type> $messageMap
```
KrzysztofCzerwinski: suggestion: `class-string`

src/Symfony/MessageMapFactory.php
```diff
@@ -0,0 +1,120 @@
...
+	}
+
+	public function create(): MessageMap
+	{
+		$returnTypesMap = [];
+
+		foreach ($this->serviceMap->getServices() as $service) {
+			$serviceClass = $service->getClass();
+
+			if (is_null($serviceClass)) {
```
KrzysztofCzerwinski: suggestion: native null comaprison- a tiny bit faster

src/Symfony/MessageMapFactory.php
```diff
@@ -0,0 +1,120 @@
...
+
+		if (count($parameters) !== 1) {
+			return;
+		}
+
+		$classNames = $parameters[0]->getType()->getObjectClassNames();
+
+		if (count($classNames) !== 1) {
+			return;
+		}
```
KrzysztofCzerwinski, line 130: suggestion: `1 !== count(...)`- not exactly a case of yoda, but seems a bit more in yoda nature where the constant value is on the left hand side

src/Type/Symfony/MessengerHandleTraitReturnTypeExtension.php
```diff
@@ -0,0 +1,85 @@
...
+		return null;
+	}
+
+	private function getMessageMap(): MessageMap
+	{
+		if (is_null($this->messageMap)) {
+			$this->messageMap = $this->messageMapFactory->create();
+		}
+
+		return $this->messageMap;
```
KrzysztofCzerwinski: suggestion: either native comparison or:
```php
$this->messageMap = $this->messageMap ?? $this->messageMapFactory->create();

return $this->messageMap;
```

src/Symfony/MessageMap.php
bnowak: thanks, adjusted

src/Symfony/MessageMap.php
bnowak: thanks, adjusted

src/Symfony/MessageMapFactory.php
bnowak: thanks, adjusted

src/Symfony/MessageMapFactory.php
bnowak: adjusted, however btw. real yoda cases are disabled in this repo in coding standards :wink:

src/Type/Symfony/MessengerHandleTraitReturnTypeExtension.php
bnowak: thanks, adjusted

KrzysztofCzerwinski (approved):

bnowak:
@ondrejmirtes any chances to review that, please? :pray:

bnowak:
@ondrejmirtes I can see that version 2.0.0 is released now. Is 1.x abandoned from now? Should I rebase that PR to 2.x branch? Any chances to review and/or merge it please? :slightly_smiling_face:

.gitignore
```diff
@@ -1,5 +1,6 @@
 /tests/tmp
 /build-cs
 /vendor
+/.idea
```
ondrejmirtes: Remove this please. This belongs to a global .gitignore.

src/Symfony/MessageMapFactory.php
```diff
@@ -0,0 +1,143 @@
...
+			}
+
+			foreach ($service->getTags() as $tag) {
+				if ($tag->getName() !== self::MESSENGER_HANDLER_TAG) {
+					continue;
+				}
+
+				/** @var array{handles?: class-string, method?: string} $tagAttributes */
+				$tagAttributes = $tag->getAttributes();
+				$reflectionClass = $this->reflectionProvider->getClass($serviceClass);
```
ondrejmirtes: This is unsafe. Ask for hasClass first.

src/Symfony/MessageMapFactory.php
```diff
@@ -0,0 +1,143 @@
...
+
+				if (isset($tagAttributes['handles'])) {
+					$handles = [$tagAttributes['handles'] => ['method' => $tagAttributes['method'] ?? self::DEFAULT_HANDLER_METHOD]];
+				} else {
+					$handles = $this->guessHandledMessages($reflectionClass);
+				}
+
+				foreach ($handles as $messageClassName => $options) {
+					$methodReflection = $reflectionClass->getNativeMethod($options['method'] ?? self::DEFAULT_HANDLER_METHOD);
```
ondrejmirtes: This is unsafe. Ask for hasNativeMethod first.

src/Symfony/MessageMapFactory.php
```diff
@@ -0,0 +1,143 @@
...
+		}
+
+		return new MessageMap($messageMap);
+	}
+
+	/** @return array<class-string, array<string, string>> */
+	private function guessHandledMessages(ClassReflection $reflectionClass): iterable
+	{
+		if ($reflectionClass->implementsInterface(MessageSubscriberInterface::class)) {
+			foreach ($reflectionClass->getName()::getHandledMessages() as $index => $value) {
```
ondrejmirtes: Save the class name into a variable first, then call the method.

src/Symfony/MessageMapFactory.php
```diff
@@ -0,0 +1,143 @@
...
+				continue;
+			}
+
+			$messageMap[$messageClassName] = $returnTypes[0];
+		}
+
+		return new MessageMap($messageMap);
+	}
+
+	/** @return array<class-string, array<string, string>> */
```
ondrejmirtes: This does not return an array. It's a generator.

src/Symfony/MessageMapFactory.php
```diff
@@ -0,0 +1,143 @@
...
+					yield $value => ['method' => self::DEFAULT_HANDLER_METHOD];
+				}
+			}
+
+			return;
+		}
+
+		try {
+			$methodReflection = $reflectionClass->getNativeMethod(self::DEFAULT_HANDLER_METHOD);
+		} catch (MissingMethodFromReflectionException $e) {
```
ondrejmirtes: Ask for hasNativeMethod first, don't catch the exception.

src/Symfony/MessageMapFactory.php
```diff
@@ -0,0 +1,143 @@
...
+			return;
+		}
+
+		$parameters = $variants[0]->getParameters();
+
+		if (count($parameters) !== 1) {
+			return;
+		}
+
+		/** @var class-string[] $classNames */
```
ondrejmirtes: Don't add this `@var`, this assumption might be wrong.

src/Symfony/MessageMapFactory.php
```diff
@@ -0,0 +1,143 @@
...
+	 */
+	private static function containOptions($index, $value): bool
+	{
+		if (is_string($index) && class_exists($index) && is_array($value)) {
+			return true;
+		} elseif (is_int($index) && is_string($value) && class_exists($value)) {
+			return false;
+		}
+
+		throw new ShouldNotHappenException();
```
ondrejmirtes: I'd say this can happen very easily. Wrong getHandledMessages implementation should not crash PHPStan.

src/Symfony/MessageMapFactory.php
```diff
@@ -0,0 +1,143 @@
...
+				continue;
+			}
+
+			foreach ($service->getTags() as $tag) {
+				if ($tag->getName() !== self::MESSENGER_HANDLER_TAG) {
+					continue;
+				}
+
+				/** @var array{handles?: class-string, method?: string} $tagAttributes */
+				$tagAttributes = $tag->getAttributes();
```
ondrejmirtes, line 55: Can we add a `@return` stub for this method instead of `@var`?

src/Symfony/MessageMapFactory.php
```diff
@@ -0,0 +1,143 @@
...
+		yield $classNames[0] => ['method' => self::DEFAULT_HANDLER_METHOD];
+	}
+
+	/**
+	 * @param mixed $index
+	 * @param mixed $value
+	 * @phpstan-assert-if-true class-string $index
+	 * @phpstan-assert-if-true array<string, mixed> $value
+	 * @phpstan-assert-if-false int $index
+	 * @phpstan-assert-if-false class-string $value
```
ondrejmirtes: All of these asserts need `=`: https://phpstan.org/writing-php-code/narrowing-types#equality-assertions

src/Type/Symfony/MessengerHandleTraitReturnTypeExtension.php
```diff
@@ -0,0 +1,86 @@
...
+	{
+		$this->messageMapFactory = $symfonyMessageMapFactory;
+	}
+
+	public function getType(Expr $expr, Scope $scope): ?Type
+	{
+		if ($this->isSupported($expr, $scope)) {
+			$arg = $expr->getArgs()[0]->value;
+			/** @var class-string[] $argClassNames */
+			$argClassNames = $scope->getType($arg)->getObjectClassNames();
```
ondrejmirtes, line 42: Don't add this `@var`.

src/Type/Symfony/MessengerHandleTraitReturnTypeExtension.php
```diff
@@ -0,0 +1,86 @@
...
+	{
+		if ($this->messageMap === null) {
+			$this->messageMap = $this->messageMapFactory->create();
+		}
+
+		return $this->messageMap;
+	}
+
+	/**
+	 * @phpstan-assert-if-true MethodCall $expr
```
ondrejmirtes: This needs `=`.

src/Type/Symfony/MessengerHandleTraitReturnTypeExtension.php
```diff
@@ -0,0 +1,86 @@
...
+			return false;
+		}
+
+		try {
+			$methodReflection = $scope->getClassReflection()->getNativeReflection()->getMethod(self::TRAIT_METHOD_NAME);
+			$declaringClassReflection = $methodReflection->getBetterReflection()->getDeclaringClass();
+
+			return $declaringClassReflection->getName() === self::TRAIT_NAME;
+		} catch (ReflectionException $e) {
+			return false;
```
ondrejmirtes, line 82: Instead of catching exception, check if the method exists first.

ondrejmirtes (changes requested):
No need to rebase this, we could release this in 1.x.

src/Symfony/MessageMapFactory.php
bnowak: right, changed to iterable :)

src/Symfony/MessageMapFactory.php
bnowak: Right, so now I'm catching that case and skipping when wrongly implemented method.

src/Symfony/MessageMapFactory.php
bnowak: This shape is dynamic and rely on what tag we're using from SF config. In this case I'm ensuring above that we're handling only `messenger.message_handler`, so we know what shape it should have.

As far I know stubs are static only, so unfortunately we cannot use them here.

Do you have other idea or it could stay as it is?

bnowak:
@ondrejmirtes Thanks for review. I addressed all your comments, please take a look again when you have time :)

Also, could you take a look on https://github.com/phpstan/phpstan-symfony/pull/404#issuecomment-2419761831 and answer that if you have an idea how to solve it please? That's my initial goal to resolve by this PR :)

src/Symfony/MessageMapFactory.php
```diff
@@ -0,0 +1,158 @@
...
+			$className = $reflectionClass->getName();
+
+			foreach ($className::getHandledMessages() as $index => $value) {
+				try {
+					if (self::containOptions($index, $value)) {
+						yield $index => $value;
+					} else {
+						yield $value => ['method' => self::DEFAULT_HANDLER_METHOD];
+					}
+				} catch (ShouldNotHappenException $e) {
```
ondrejmirtes: Never catch ShouldNotHappenException.

src/Type/Symfony/MessengerHandleTraitReturnTypeExtension.php
```diff
@@ -0,0 +1,86 @@
...
+
+	public function __construct(MessageMapFactory $symfonyMessageMapFactory)
+	{
+		$this->messageMapFactory = $symfonyMessageMapFactory;
+	}
+
+	public function getType(Expr $expr, Scope $scope): ?Type
+	{
+		if ($this->isSupported($expr, $scope)) {
+			$arg = $expr->getArgs()[0]->value;
```
ondrejmirtes: `$expr->getArgs()[0]` might not exist. Check the existence first.

ondrejmirtes (changes requested):

bnowak:
@ondrejmirtes both comments fixed. Could you re-check? Thanks, in advance :)

ondrejmirtes:
Thank you. Please be available if any issues arise.

bnowak:
@ondrejmirtes Thank you for merging and your help with reviewing that :)

Could you take a look on https://github.com/phpstan/phpstan-symfony/pull/404#issuecomment-2419761831 comment please 🙏 I'm not sure whether I should create separate issue for that? Maybe it's able to achive with already existing features/annotations? If not, do you have an idea how to implement such feature?

ondrejmirtes:
@bnowak Feel free to send a failing test so I can understand it better.

ondrejmirtes:
Hi, since MessageSubscriberInterface has been deprecated/removed in favour of `#[AsMessageHandler]`, shouldn't the code in MessageMapFactory be adjusted to also support the attribute? Thanks.

bnowak:
@ondrejmirtes I'll take a look on that tomorrow. Thanks for info :wink:

ondrejmirtes:
Maybe it's the reason why some of the tests are failing on Symfony 7.3, maybe not: https://github.com/phpstan/phpstan-symfony/actions/runs/16193692378/job/45715050598?pr=447

bnowak:
Yes it is. I started working on tests separation and skipping the case when `MessageSubscriberInterface` is missing (SF <7.0). Tomorrow, I'll push PR for that.

Regarding adding support for `#[AsMessageHandler]` - I'll handle that in separate PR, no problem. However, I think firstly before that, it would be nice to have support for SF 7.3 in tests (work which you started in https://github.com/phpstan/phpstan-symfony/pull/447/files). Then, I'd add that support.

@ondrejmirtes What do you think? Does such order make sense?

ondrejmirtes:
I think AsMessageHandler already existed before Symfony 7, didn't it?

It'd make it easy to implement it here, and when my symfony-7-3 is rebased, less tests should fail.

bnowak:
Ahh, you're right. I don't know why I assumed it was introduced in 7.0 😅

Tomorrow will try to handle both topics 😉

bnowak:
@ondrejmirtes please review https://github.com/phpstan/phpstan-symfony/pull/448.

Regarding `AsMessageHandler` support - it's just tag for auto configuring it during kernel boot, so its result is generated container xml file definition which this plugin read from (the same as messenger handle trait feature).
Given that, it should be supported by default and additional tests are not needed IMO.

PR #405, opened 2024-08-21
Add Twig template exists rule

### LINKED ISSUE #257: verify that the file in render actually exists
A common error we have is "you forgot to commit the  .html.twig file" or the twig file was renamed and not all occurences was replaced.

If phpstan could detect the path is linked to a missing file, that would help a lot :)

zacharylund:
Fixes #257


tests/Rules/Symfony/ExampleTwigController.php
```diff
@@ -0,0 +1,71 @@
...
+		$this->renderForm('foo.html.twig');
+		$this->renderView('foo.html.twig');
+		$this->stream('foo.html.twig');
+
+		$this->render('bar.html.twig');
+		$this->renderBlock('bar.html.twig');
+		$this->renderBlockView('bar.html.twig');
+		$this->renderForm('bar.html.twig');
+		$this->renderView('bar.html.twig');
+		$this->stream('bar.html.twig');
```
ruudk, line 27: How to deal with namespaces like `@Admin/index.html.twig`? 
https://twig.symfony.com/doc/3.x/api.html#twig-loader-filesystemloader

zacharylund: One option would be to duplicate Symfony's config in PHPStan's config:

https://symfony.com/doc/current/reference/configuration/twig.html#paths

The other option (and what I considered initially) would be to instantiate a Twig loader much the same way we instantiate a Symfony console application.

zacharylund: @ruudk I've updated this to handle Twig namespaces.

extension.neon
```diff
@@ -115,6 +116,7 @@ parametersSchema:
 		constantHassers: bool()
 		console_application_loader: schema(string(), nullable())
```
ruudk, line 117: Why this snake_case syntax

ruudk (commented):
I would use the Twig environment using a loader, then it auto picks up all namespaces.

extension.neon
zacharylund: This exists for backward compatibility and is out of scope for this PR. It's also available as `consoleApplicationLoader`.

zacharylund:
> I would use the Twig environment using a loader, then it auto picks up all namespaces.

@ruudk, I've updated this to use a loader, but I'm not loving it. Because `twig` is a private service, the loader would need to look something like:

```php
use App\Kernel;

use App\Kernel;
use Symfony\Component\Dotenv\Dotenv;

require __DIR__ . '/../vendor/autoload.php';

(new Dotenv())->bootEnv(__DIR__ . '/../.env');

$kernel = new Kernel('test', (bool) $_SERVER['APP_DEBUG']);
$kernel->boot();

return $kernel->getContainer()->get('test.service_container')->get('twig');
```

ruudk:
Yes. Downside is that now you are using the `test` environment.

In [TwigStan](https://github.com/twigstan/twigstan) I came up with this solution: 
https://github.com/twigstan/twigstan/blob/90e1792a68dc0cd9471ce87a1e1f0145467299f9/twig-loader-symfony.php.dist#L18-L26

src/Rules/Symfony/TwigTemplateExistsRule.php
```diff
@@ -0,0 +1,104 @@
...
+
+		if ($templateArg->value instanceof Variable && is_string($templateArg->value->name)) {
+			$varType = $scope->getVariableType($templateArg->value->name);
+
+			foreach ($varType->getConstantStrings() as $constantString) {
+				$templateNames[] = $constantString->getValue();
+			}
+		} elseif ($templateArg->value instanceof String_) {
+			$templateNames[] = $templateArg->value->value;
+		}
```
VincentLanglet, line 58: I think you can avoid the if/else and write
```suggestion
			$argType = $scope->getType($templateArg->value);

			foreach ($varType->getConstantStrings() as $constantString) {
				$templateNames[] = $constantString->getValue();
			}
```

src/Rules/Symfony/TwigTemplateExistsRule.php
```diff
@@ -0,0 +1,104 @@
...
+		foreach ($templateNames as $templateName) {
+			if ($this->twigEnvironmentResolver->templateExists($templateName)) {
+				continue;
+			}
+
+			$errors[] = RuleErrorBuilder::message(sprintf(
+				'Twig template "%s" does not exist.',
+				$templateName
+			))->line($templateArg->getStartLine())->identifier('twig.templateNotFound')->build();
+		}
```
VincentLanglet, line 75: I dunno if the error strategy need to depends on the PHPStan Level.

For instance, maybe `existingTemplate|nonExistingTemplate` should only be reported at level 7 or more.
In the same way unions/maybe are reported on level 7 for classic rules.

PR #408, opened 2024-09-17
Fix the KernelTestCase stub

stof:
KernelTestCase does not have a `$container` static property


ondrejmirtes:
It's there on older Symfony versions.

stof:
@ondrejmirtes but this causes issues when using newer versions of Symfony, because phpstan will think that the property is still there.

ondrejmirtes:
You can’t conjure things that don’t exist into existence with stub files in PHPStan.

It’s just for overriding PHPDocs.

PR #409, opened 2024-09-26
Do not use InputBagTypeSpecifyingExtension on true context.

VincentLanglet:
The `InputBag::has` method rely on `array_key_exists`

see
https://github.com/symfony/symfony/blob/e7589c18012314ef262f040357b832a264b2a6eb/src/Symfony/Component/HttpFoundation/ParameterBag.php#L86-L89

So the following code is valid
```
$inputBag->set('foo', null);
$inputBag->has('foo'); // true
$inputBag->get('foo'); // null
```

:/


ondrejmirtes:
Thank you.

Issue #417, opened 2024-12-04
Missing stub file for Symfony Form DataMapperInterface


roerbakei:
### Bug Report / Feature Request

When implementing the [Symfony Form DataMapperInterface](https://symfony.com/doc/current/form/data_mappers.html) it will currently throw the following error:

```php
Method  MyApp\FormType::mapDataToForms()  
         has parameter $forms with generic interface                                                                               
         Symfony\Component\Form\FormInterface but does not specify its types:                                                      
         TData     
 ```
 
 When fixing this by defining TData for the FormInterface in the class that implements the DataMapperInterface, the type will be different from the interface resulting in an error:
 
 ```php
    /**
     * @param Traversable<FormInterface<mixed>> $forms
     */
    public function mapDataToForms(mixed $viewData, Traversable $forms): void
```

> Note: I've typed TData of FormInterface to mixed since `$forms` stands for the child forms of the form which is using the data mapper and so can have multiple types.

```php
Parameter #2 $forms (Traversable<mixed,                                                                                   
         Symfony\Component\Form\FormInterface<mixed>>) of method                                                                   
         MyApp\FormType::mapDataToForms()  
         should be contravariant with parameter $forms (Traversable<mixed,                                                         
         Symfony\Component\Form\FormInterface>) of method                                                                          
         Symfony\Component\Form\DataMapperInterface::mapDataToForms()           
 ```
 
 Unless I'm missing something it looks that a stub for Symfony Form DataMapperInterface is not provided yet.

mmarton:
I have the same issue

JefvdA:
@mmarton you can temporarly resolve your issue by using the stub file @roerbakei provided in his PR -> https://github.com/phpstan/phpstan-symfony/pull/418/files

[PHPStan docs](https://phpstan.org/user-guide/stub-files) explain how to add local stub-files

JefvdA:
Created a PR with @roerbakei his suggested change ( adding the stub file for the `DataItemMapperInterface` ), but with a small tweak that fixes the feedback I gave ( https://github.com/phpstan/phpstan-symfony/pull/418#discussion_r1905303708 )

PR #418, opened 2024-12-04
feat: stub DataMapperInterface

roerbakei:
- Ensure TData of FormInterface is provided.
- Set TData on mixed since we don't know the type of the child forms. 
- Add template for DataMapperInterface.

I've set TData of FormInterface to mixed since each child form can have a different type.

refs: #417


stubs/Symfony/Component/Form/DataMapperInterface.stub
```diff
@@ -0,0 +1,22 @@
...
+     * @param \Traversable<mixed, FormInterface<mixed>> $forms
+     */
+    public function mapDataToForms(mixed $viewData, \Traversable $forms): void;
+
+    /**
+     * @param \Traversable<mixed, FormInterface<mixed>> $forms
+     * @param TViewData $viewData
+     * @param-out TViewData $viewData
+     */
+    public function mapFormsToData(\Traversable $forms, mixed &$viewData): void;
```
JefvdA, line 21: As it's stated in the docs ( https://symfony.com/doc/current/form/data_mappers.html#creating-a-data-mapper ), the following logic is needed:

```php
// there is no data yet, so nothing to prepopulate
if (null === $viewData) {
    return;
}
```

This, as it seems that `mapDataToForms` is called before the data is avaialble. Which will mean that `$viewData` will be `null` at that time.

---

Thus I propose the following change:

```suggestion
     * @param TViewData|null $viewData
     * @param \Traversable<mixed, FormInterface<mixed>> $forms
     */
    public function mapDataToForms(mixed $viewData, \Traversable $forms): void;

    /**
     * @param \Traversable<mixed, FormInterface<mixed>> $forms
     * @param TViewData|null $viewData
     * @param-out TViewData $viewData
     */
    public function mapFormsToData(\Traversable $forms, mixed &$viewData): void;
```

---

For example, [the stub file for the `FormTypeInterface`](https://github.com/phpstan/phpstan-symfony/blob/c08cd8e54a08d651bc402d304cfa161c3c3766c4/stubs/Symfony/Component/Form/FormTypeInterface.stub#L11) also has a generic type, but also has it defined nullable when it's being passed to it's `buildForm` method.

ondrejmirtes:
There are two different PRs with this stub: https://github.com/phpstan/phpstan-symfony/pull/422

1) Both have the same mistake of not registering the stub.
2) Both are slightly different so you should agree first how it should look like.

PR #419, opened 2024-12-24
Add stub for form options

VincentLanglet:


tests/Type/Symfony/data/form_options.php
```diff
@@ -0,0 +1,102 @@
...
+
+	public function doSomethingWithOption(): void
+	{
+		$form = $this->formFactory->create(DataClassType::class, new DataClass(), ['required' => 'foo']);
+		assertType('Symfony\Component\Form\FormInterface<GenericFormOptionsType\DataClass>', $form);
+	}
+
+	public function doSomethingWithInvalidOption(): void
+	{
+		$form = $this->formFactory->create(DataClassType::class, new DataClass(), ['required' => 42]);
```
VincentLanglet, line 75: I'd like to get an error here saying I pass an integer when a string is expected because DataClassType is
```
AbstractType<DataClass, array{required: string, optional: int}>
```

I thought it would have been reported by the CallMethodsRuleTest.

What do I miss @ondrejmirtes ?

ondrejmirtes: Please recreate the scenario in a minimal way on phpstan.org/try so I don't have to hold all the interfaces and Symfony classes in my head.

stubs/Symfony/Component/Form/AbstractType.stub
```diff
@@ -4,27 +4,28 @@ namespace Symfony\Component\Form;
 
 /**
  * @template TData
+ * @template TOptions of array<string, mixed>
```
ondrejmirtes, line 7: This has to be optional, otherwise it's a BC break. See https://github.com/phpstan/phpstan-src/pull/3457

ondrejmirtes (changes requested):

tests/Type/Symfony/data/form_options.php
VincentLanglet: Sure.
You already helped me with https://github.com/phpstan/phpstan/discussions/12372#discussioncomment-11881550.

But I discovered that in Symfony, that the required options in the Factory::create methods and the certains options in the buildForm $options are not exactly the same.

So this would give something like
https://phpstan.org/r/2fc552e7-506e-46a5-aba0-69287a5b8d3e

I dunno if it's possible to have something better than
```
@implements FormTypeInterface<
 *   object,
 *   array{required: string, optional?: int, defined?: bool},
 *   array{required: string, optional: int, defined?: bool}
 * >
```

PR #420, opened 2024-12-30
Support Symfony #[AutowireLocator] attribute

### LINKED ISSUE #411: Support `#[AutowireLocator]`
Symfony 6.4 introduced an [attribute called `#[AutowireLocator]`](https://symfony.com/doc/6.4/service_container/service_subscribers_locators.html#the-autowirelocator-and-autowireiterator-attributes) which automatically creates and injects a `ServiceLocator` into a service. Looks like a similar feature was already implemented with #151.

Is it possible to add support for this feature, too?

```php
final class TranscriptAdmin
{
   // ...

    #[Required]
    public function setContainer(
        #[AutowireLocator([ProfileRepository::class])] ContainerInterface $container
    ): void {
        $this->locator = $container;
    }
}
```

```
 ------ ---------------------------------------------------------------------- 
  Line   src/TranscriptBundle/Admin/TranscriptAdmin.php                        
 ------ ---------------------------------------------------------------------- 
  56     Service "App\ProfileBundle\Repository\ProfileRepository" is private.  
         ✏️  src/TranscriptBundle/Admin/TranscriptAdmin.php                    
 ------ ---------------------------------------------------------------------- 
```

RafaelKr:
I implemented initial support for the `#[AutowireLocator]` attribute. See https://symfony.com/blog/new-in-symfony-6-4-autowirelocator-and-autowireiterator-attributes

The simple syntax is supported now:
```php
#[AutowireLocator([FooHandler::class, BarHandler::class])]
```

It also supports when classes are specified via string:
```php
#[AutowireLocator(['App\FooHandler'])]
```

The advanced syntax described in the blog post is ~~not~~ also supported ~~yet~~ now:
```php
#[AutowireLocator([
    'foo' => FooHandler::class,
    'bar' => new SubscribedService(type: 'string', attributes: new Autowire('%some.parameter%')),
    'optionalBaz' => '?'.BazHandler::class,
])]
private ContainerInterface $handlers,
```

~~Maybe someone else wants to figure this out. Unfortunately I have no more time left to do it in the coming days.~~

Edit 2: I just found out how I can both simplify this a lot and implement the advanced syntax by just creating an instance of the AutowireLocator attribute. And I'll move this to `AutowireLoaderServiceMapFactory` so it will be also available for `PHPStan\Rules\Symfony\ContainerInterfaceUnknownServiceRule`.

closes #411

### DIFF
--- a/src/Rules/Symfony/ContainerInterfacePrivateServiceRule.php
+++ b/src/Rules/Symfony/ContainerInterfacePrivateServiceRule.php
@@ -5,12 +5,19 @@ namespace PHPStan\Rules\Symfony;
 use PhpParser\Node;
 use PhpParser\Node\Expr\MethodCall;
 use PHPStan\Analyser\Scope;
+use PHPStan\BetterReflection\Reflection\Adapter\FakeReflectionAttribute;
+use PHPStan\BetterReflection\Reflection\Adapter\ReflectionAttribute;
+use PHPStan\Reflection\ClassReflection;
 use PHPStan\Rules\Rule;
 use PHPStan\Rules\RuleErrorBuilder;
+use PHPStan\Symfony\ServiceDefinition;
 use PHPStan\Symfony\ServiceMap;
 use PHPStan\TrinaryLogic;
 use PHPStan\Type\ObjectType;
 use PHPStan\Type\Type;
+use Symfony\Component\DependencyInjection\Attribute\AutowireLocator;
+use function class_exists;
+use function get_class;
 use function sprintf;
 
 /**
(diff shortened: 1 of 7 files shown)


RafaelKr:
~~I tried to fix all errors but I'm not sure how to fix those 2 errors left.~~

Edit: Got it.

RafaelKr:
I'll add some tests later, besides from that it's ready for review.

Btw: I tried to rename the branch of my Repo to feat/autowirelocator, but it wasn't possible without losing the connection to this PR, the PR showed the branch was deleted in that case.

RafaelKr:
I also added tests now, but I have no idea how to resolve the failing lint and test checks. If someone can resolve those we're ready to merge from my side :rocket:

RafaelKr:
I also noticed it only works when we're using [Constructor property promotion](https://stitcher.io/blog/constructor-promotion-in-php-8), so when specifying `private ContainerInterface $locator` in the constructor.

It won't work when `$locator` is specified as a class member and then assigned in the constructor method (as here e.g. https://github.com/phpstan/phpstan-symfony/blob/c7b7e7f520893621558bfbfdb2694d4364565c1d/tests/Rules/Symfony/ExampleServiceSubscriber.php).

Is this a problem in practice?

ondrejmirtes:
Hi, I can't review this when the build isn't green.

PR #421, opened 2025-01-07
Add result cache meta extension for DI container

### LINKED ISSUE #255: Result cache is not invalidated after DI container dump
Imagine scenario:

- With `bin/console cache:clear` you dump DI container
- In code you use `$container->get('foo');`
- PHPStan reports `Service "foo" is not registered in the container`
- You fix DI definition by adding `foo` service
- `cache:clear` again, so DI definition contains `foo` service
- Because of result cache PHPStan reports errors even though DI container definition has changed and service is registered

Expected result: refreshing DI container invalidates result cache for files that uses DI (have errors reported by Symfony extension).

Wirone:
> [!IMPORTANT]
> This feature was brought to you thanks to [GetResponse](https://getresponse.com) - Marketing Software for Professional Email Marketing, which paid for my work on this.

Fixes phpstan/phpstan-symfony#255, continues phpstan/phpstan-src#3765.

---

Following [this experiment's results](https://github.com/phpstan/phpstan-symfony/issues/255#issuecomment-2373929376) I decided to calculate meta hash based on DI container's actual content (params and services), without interacting with XML file directly. `ParameterMapFactory` and `ServiceMapFactory` were used for retrieving data, also I implemented "singleton" approach there, so the XML file won't be processed multiple times (once for calculating hash, then for processing rules).

I start with a draft PR so I can get early feedback. I also need help when it comes to testing this, because I don't see e2e approach here. I wanted to test it on our codebase, but it's really cumbersome to setup Composer dependencies to get dev version of PHPStan and the plugin 😅.

### DIFF
--- a/extension.neon
+++ b/extension.neon
@@ -363,3 +363,8 @@ services:
 	-
 		factory: PHPStan\Type\Symfony\ExtensionGetConfigurationReturnTypeExtension
 		tags: [phpstan.broker.dynamicMethodReturnTypeExtension]
+
+	-
+		class: PHPStan\Symfony\SymfonyContainerResultCacheMetaExtension
+		tags:
+			- phpstan.resultCacheMetaExtension
--- a/src/Symfony/SymfonyContainerResultCacheMetaExtension.php
+++ b/src/Symfony/SymfonyContainerResultCacheMetaExtension.php
@@ -1,0 +1,54 @@
+<?php declare(strict_types = 1);
+
+namespace PHPStan\Symfony;
+
+use PHPStan\Analyser\ResultCache\ResultCacheMetaExtension;
+use function array_map;
+use function hash;
+use function serialize;
+
+final class SymfonyContainerResultCacheMetaExtension implements ResultCacheMetaExtension
+{
+
+	private ParameterMapFactory $parameterMapFactory;
+
+	private ServiceMapFactory $serviceMapFactory;
+
+	public function __construct(
+		ParameterMapFactory $parameterMapFactory,
+		ServiceMapFactory $serviceMapFactory
+	)
+	{
+		$this->parameterMapFactory = $parameterMapFactory;
+		$this->serviceMapFactory = $serviceMapFactory;
+	}
+
+	public function getKey(): string
+	{
+		return 'symfonyDiContainer';
+	}
+
+	public function getHash(): string
+	{
+		return hash('sha256', serialize([
+			'parameters' => array_map(
+				static fn (ParameterDefinition $parameter) => [
+					'name' => $parameter->getKey(),
+					'value' => $parameter->getValue(),
+				],
+				$this->parameterMapFactory->create()->getParameters(),
+			),
+			'services' => array_map(
+				static fn (ServiceDefinition $service) => [
+					'id' => $service->getId(),
+					'class' => $service->getClass(),
+					'public' => $service->isPublic() ? 'yes' : 'no',
+					'synthetic' => $service->isSynthetic() ? 'yes' : 'no',
+					'alias' => $service->getAlias(),
+				],
+				$this->serviceMapFactory->create()->getServices(),
+			),
+		]));
+	}
+
+}
--- a/src/Symfony/SymfonyContainerResultCacheMetaExtension.php
+++ b/src/Symfony/SymfonyContainerResultCacheMetaExtension.php
@@ -10,17 +10,14 @@ use function serialize;
 final class SymfonyContainerResultCacheMetaExtension implements ResultCacheMetaExtension
 {
 
-	private ParameterMapFactory $parameterMapFactory;
+	private ParameterMap $parameterMap;
 
-	private ServiceMapFactory $serviceMapFactory;
+	private ServiceMap $serviceMap;
 
-	public function __construct(
-		ParameterMapFactory $parameterMapFactory,
-		ServiceMapFactory $serviceMapFactory
-	)
+	public function __construct(ParameterMap $parameterMap, ServiceMap $serviceMap)
 	{
-		$this->parameterMapFactory = $parameterMapFactory;
-		$this->serviceMapFactory = $serviceMapFactory;
+		$this->parameterMap = $parameterMap;
+		$this->serviceMap = $serviceMap;
 	}
 
 	public function getKey(): string
@@ -36,7 +33,7 @@ final class SymfonyContainerResultCacheMetaExtension implements ResultCacheMetaE
 					'name' => $parameter->getKey(),
 					'value' => $parameter->getValue(),
 				],
-				$this->parameterMapFactory->create()->getParameters(),
+				$this->parameterMap->getParameters(),
 			),
 			'services' => array_map(
 				static fn (ServiceDefinition $service) => [
@@ -46,7 +43,7 @@ final class SymfonyContainerResultCacheMetaExtension implements ResultCacheMetaE
 					'synthetic' => $service->isSynthetic() ? 'yes' : 'no',
 					'alias' => $service->getAlias(),
 				],
-				$this->serviceMapFactory->create()->getServices(),
+				$this->serviceMap->getServices(),
 			),
 		]));
 	}
--- a/src/Symfony/SymfonyContainerResultCacheMetaExtension.php
+++ b/src/Symfony/SymfonyContainerResultCacheMetaExtension.php
@@ -6,6 +6,7 @@ use PHPStan\Analyser\ResultCache\ResultCacheMetaExtension;
 use function array_map;
 use function hash;
 use function serialize;
+use function sort;
 
 final class SymfonyContainerResultCacheMetaExtension implements ResultCacheMetaExtension
 {
@@ -29,20 +30,32 @@ final class SymfonyContainerResultCacheMetaExtension implements ResultCacheMetaE
 	{
 		return hash('sha256', serialize([
 			'parameters' => array_map(
-				static fn (ParameterDefinition $parameter) => [
+				static fn (ParameterDefinition $parameter): array => [
 					'name' => $parameter->getKey(),
 					'value' => $parameter->getValue(),
 				],
 				$this->parameterMap->getParameters(),
 			),
 			'services' => array_map(
-				static fn (ServiceDefinition $service) => [
-					'id' => $service->getId(),
-					'class' => $service->getClass(),
-					'public' => $service->isPublic() ? 'yes' : 'no',
-					'synthetic' => $service->isSynthetic() ? 'yes' : 'no',
-					'alias' => $service->getAlias(),
-				],
+				static function (ServiceDefinition $service): array {
+					$serviceTags = array_map(
+						static fn (ServiceTag $tag) => [
+							'name' => $tag->getName(),
+							'attributes' => $tag->getAttributes(),
+						],
+						$service->getTags(),
+					);
+					sort($serviceTags);
+
+					return [
+						'id' => $service->getId(),
+						'class' => $service->getClass(),
+						'public' => $service->isPublic() ? 'yes' : 'no',
+						'synthetic' => $service->isSynthetic() ? 'yes' : 'no',
+						'alias' => $service->getAlias(),
+						'tags' => $serviceTags,
+					];
+				},
 				$this->serviceMap->getServices(),
 			),
 		]));
--- a/src/Symfony/SymfonyContainerResultCacheMetaExtension.php
+++ b/src/Symfony/SymfonyContainerResultCacheMetaExtension.php
@@ -6,8 +6,8 @@ use PHPStan\Analyser\ResultCache\ResultCacheMetaExtension;
 use function array_map;
 use function hash;
 use function ksort;
-use function serialize;
 use function sort;
+use function var_export;
 
 final class SymfonyContainerResultCacheMetaExtension implements ResultCacheMetaExtension
 {
@@ -56,7 +56,7 @@ final class SymfonyContainerResultCacheMetaExtension implements ResultCacheMetaE
 		}
 		ksort($services);
 
-		return hash('sha256', serialize(['parameters' => $parameters, 'services' => $services]));
+		return hash('sha256', var_export(['parameters' => $parameters, 'services' => $services], true));
 	}
 
 }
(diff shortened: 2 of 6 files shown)


Wirone:
@ondrejmirtes did you have opportunity to look at it? I hope for your feedback as in the main extension point PR, so I can finish this properly 🙂.

ondrejmirtes:
It's a bit crazy to urge me less than 24h after you opened the PR. Please avoid that next time.

A literally wasn't in front of my computer since then.

Wirone:
Sorry, I just didn't know you were notified about the PR because I realised I did not mention you in the description 😅. Sorry for bothering, take your time of course, just wanted to know if you're aware about this one.

src/Symfony/SymfonyContainerResultCacheMetaExtension.php
```diff
@@ -0,0 +1,54 @@
...
+	}
+
+	public function getKey(): string
+	{
+		return 'symfonyDiContainer';
+	}
+
+	public function getHash(): string
+	{
+		return hash('sha256', serialize([
```
ondrejmirtes: I'm not a huge fan of serialization like this. Something unexpected might occur there, some values that's not possible to serialize. What about just hashing the contents of the XML?

src/Symfony/XmlServiceMapFactory.php
```diff
@@ -85,7 +92,9 @@ public function create(): ServiceMap
 			);
 		}
 
-		return new DefaultServiceMap($services);
+		ksort($services);
+
+		return $this->serviceMap = new DefaultServiceMap($services);
```
ondrejmirtes: This doesn't make sense. This is a stateless service, it shouldn't be cached. We're not going to be asking for these objects in the extension anyway, so you can throw this away :)

ondrejmirtes (changes requested):
You can test this by:

1) Create `SymfonyContainerResultCacheMetaExtension extends PHPStanTestCase`
2) Create a new .neon file under tests/ and link to an actual dumped container XML file which you're going to commit in this repository under tests/.
3) Override `getAdditionalConfigFiles` method and supply path to the .neon file from 2)
4) Create `new SymfonyContainerResultCacheMetaExtension` and supply constructor arguments via `self::getContainer()->getByType()`

src/Symfony/SymfonyContainerResultCacheMetaExtension.php
Wirone: As I wrote in the description (well, in the linked comment), interacting directly with XML has the drawback that result cache can be invalidated even if the container wasn't changed, only params/services were moved to other place of XML, as the final output depends on the order of source definition files. By calculating hash from parsed and sorted params/services we can be sure it will only be invalidated when DI definition was _actually changed_.

ondrejmirtes: Oh I see, you're already making the scalar values here. So there's not going to be any surprise like getting the actual service objects and trying to serialize them.

I'd just exchange the `serialize` for something like `var_export` - ResultCacheManager uses that in many places.

And ask for ServiceMap/ParameterMap in the constructor instead of the factories. Thanks.

src/Symfony/XmlServiceMapFactory.php
Wirone: In current implementation it makes sense as those factories are used both in the extension and other places. Unless the Nette container doesn't use the same instance for dependency injection (I don't know it, you tell me 😅) it is performance optimisation, to avoid parsing same XML files multiple times (GetResponse's DI container has 2.5MB xml file, so there's no point in parsing it multiple times, in fact it would be better if params and services were parsed in the same batch, not separately).

ondrejmirtes: I think you can just ask for ParameterMap or ServiceMap in the constructor and it will work. I did that during my SymfonyCon talk.

Wirone: I did not know it, will try it, but I believe it does not change the fact this improvement is needed. If Nette calls the factory every time `ParameterMap` or `ServiceMap` is requested on DI, it should process the file only once, then reuse the result. I will check it soon 🙂.

ondrejmirtes: As any sensible DI container, nette/di keeps the services and reuses them once created.

ondrejmirtes: It's possible that some work is done twice for parameters and services but I'm not sure about that + it should be focus of a different PR.

src/Symfony/SymfonyContainerResultCacheMetaExtension.php
Wirone: Maps are injected directly now.

src/Symfony/XmlServiceMapFactory.php
Wirone: I removed state from the factories, but kept sorting.

Wirone:
@ondrejmirtes please make sure you're familiar with the commits' messages, as I provided additional info there. When it comes to `serialize()` I believe it's more stable than `var_export()` when it comes to producing string that is used for calculating hash. We could use `json_encode()` too, but it would require adding `ext-json` to Composer's dependencies.

Again, thank you very much for providing feedback and guiding about tests 🍻.

Wirone:
To be honest, I am not sure whether services' tags should be taken into consideration when calculating container's hash 🤔. `Symfony\Component\DependencyInjection\ContainerInterface` and even `Symfony\Component\DependencyInjection\Container` doesn't expose any API for interacting with tags, these are rather used in compiler passes during container build. But since `PHPStan\Symfony\ServiceDefinition` contains tags, I included them in the calculation. Let me know what you think @ondrejmirtes 🙂.

extension.neon
```diff
@@ -8,6 +8,7 @@ parameters:
 		containerXmlPath: null
 		constantHassers: true
 		consoleApplicationLoader: null
+		invalidateResultCacheOnDiDefinitionChange: true
```
ondrejmirtes: I don't think we need an option for this.

tests/Symfony/SymfonyContainerResultCacheMetaExtensionTest.php
```diff
@@ -0,0 +1,402 @@
...
+final class SymfonyContainerResultCacheMetaExtensionTest extends PHPStanTestCase
+{
+
+	private static string $configFilePath;
+
+	/**
+	 * This test has to check if hash of the Symfony Container is correctly calculated,
+	 * in order to do that we need to dynamically create a temporary configuration file
+	 * because `PHPStanTestCase::getContainer()` caches container under key calculated from
+	 * additional config files' paths, so we can't reuse the same config file between tests.
```
ondrejmirtes: Seems hacky and I don't understand it.

tests/Symfony/SymfonyContainerResultCacheMetaExtensionTest.php
```diff
@@ -0,0 +1,402 @@
...
+	 */
+	public static function getAdditionalConfigFiles(): array
+	{
+		return [
+			__DIR__ . '/../../extension.neon',
+			self::$configFilePath,
```
ondrejmirtes: I don't understand the need for a static property here. There are always problems with them.

tests/Symfony/SymfonyContainerResultCacheMetaExtensionTest.php
```diff
@@ -0,0 +1,402 @@
...
+				XML,
+		];
+	}
+
+	private function calculateSymfonyContainerHash(
+		string $xmlContent,
+		bool $invalidateResultCacheOnDiDefinitionChange = true
+	): string
+	{
+		$symfonyContainerXmlPath = tempnam(sys_get_temp_dir(), 'phpstan-meta-extension-test-container-xml-');
```
ondrejmirtes: I don't like this dynamic approach. I also don't like the XML strings in data providers. Do we really need to test multiple XML strings? I imagined we would simply dump a single XML path to disk, test the calculated hash from that, and call it a day. It's sufficient the code would be executed once.

ondrejmirtes (changes requested):

extension.neon
Wirone: I thought it would be better to provide the toggle so people can opt-out if they find this feature somehow stepping in their way of doing things, while continuing to use Symfony extension. I can remove it if you think it's better to always have it enabled.

ondrejmirtes: Sure, leave it always enabled until we get feedback we can act on.

ondrejmirtes: I mean remove the option to disable the extension.

tests/Symfony/SymfonyContainerResultCacheMetaExtensionTest.php
Wirone: Are you familiar with [this commit's description](https://github.com/phpstan/phpstan-symfony/pull/421/commits/a561077e30a490526498a00d5f6ac47b5a254a73)? This is done like this to work around  `PHPStanTestCase::getContainer()`, which caches the container under the hash calculated from the list of additional config files. So I can't provide these paths statically, as it would lead to cache reuse for each test case (and that would load the config pointing to one specific container XML every time, while I need different container definitions to be loaded - see [here](#discussion_r1922178151)).

tests/Symfony/SymfonyContainerResultCacheMetaExtensionTest.php
Wirone: Please have a look [here](#discussion_r1922158510) for details.

tests/Symfony/SymfonyContainerResultCacheMetaExtensionTest.php
Wirone: Container's definition is exactly the same (service/parameter-wise) if XML elements change their order, or XML nodes' attributes change their order without changing values. I wanted to make sure invalidation is done _only when actually needed_, test cases cover that. This makes it more stable, as it's not strictly dependant on how Symfony dumps container or how devs structure their DI files in their projects ([my experiment](https://github.com/phpstan/phpstan-symfony/issues/255#issuecomment-2373929376) proved that moving the same services between files and changing imports lead to "different" final XML - with exact same scope, but different order of elements). I believe it's better to test actual behaviour of invalidating cache (difference, or lack of it, in calculated hashes) rather than calculated hash's value, which does not tell us if it's going to be invalidated or not.

Strings in data provider are technical detail, I can move it to files if you want - personally for me it's more readable like this, as you don't need to jump between files to analyse the flow and expected behaviour.

ondrejmirtes: This is a very complicated and brittle setup. Clearly the `self::getContainer()` thing is doing more harm than good here. Please create ParameterMap and ServiceMap by doing `new XmlParameterMapFactory()` and `XmlServiceMapFactory` respectively.

ondrejmirtes:
I will also get rid of the Configuration class in 2.0.x now.

extension.neon
Wirone: Configurability was removed.

tests/Symfony/SymfonyContainerResultCacheMetaExtensionTest.php
Wirone: Whole I/O overhead was removed.

tests/Symfony/SymfonyContainerResultCacheMetaExtensionTest.php
Wirone: Tests now create `ServiceMap` and `ParameterMap` instances instead of taking them from container.

tests/Symfony/SymfonyContainerResultCacheMetaExtensionTest.php
Wirone: I/O overhead was removed.

src/Symfony/SymfonyContainerResultCacheMetaExtension.php
```diff
@@ -0,0 +1,62 @@
...
+				'class' => $service->getClass(),
+				'public' => $service->isPublic() ? 'yes' : 'no',
+				'synthetic' => $service->isSynthetic() ? 'yes' : 'no',
+				'alias' => $service->getAlias(),
+				'tags' => $serviceTags,
+			];
+		}
+		ksort($services);
+
+		return hash('sha256', serialize(['parameters' => $parameters, 'services' => $services]));
```
ondrejmirtes: I'd rather have var_export instead of serialize

src/Symfony/XmlParameterMapFactory.php
```diff
@@ -40,18 +42,23 @@ public function create(): ParameterMap
 
 		/** @var Parameter[] $parameters */
 		$parameters = [];
-		foreach ($xml->parameters->parameter as $def) {
-			/** @var SimpleXMLElement $attrs */
-			$attrs = $def->attributes();
 
-			$parameter = new Parameter(
```
ondrejmirtes, line 47: What are these changes for? Can you remove them?

src/Symfony/XmlServiceMapFactory.php
```diff
@@ -39,35 +41,38 @@ public function create(): ServiceMap
 		$services = [];
 		/** @var Service[] $aliases */
 		$aliases = [];
-		foreach ($xml->services->service as $def) {
```
ondrejmirtes, line 42: What are these changes for? Can you remove them?

ondrejmirtes (changes requested):

src/Symfony/XmlParameterMapFactory.php
Wirone: This is a fix for handling XML files without `<services>` / `<parameters>`, I stumbled upon this when tests read container from files and I did not want to declare both services and parameters in each test case. I believe it's not needed now for _my_ usage, but it still fixes potential problem with using undefined properties. Let me know if you want to keep it or not.

Wirone: @ondrejmirtes for a wider context: if I understand [Symfony's container schema](https://symfony.com/schema/dic/services/services-1.0.xsd) both `parameters` and `services` nodes may occur 0 or 1 times (`minOccurs="0"` without `maxOccurs`), so it's not only a theoretical fix, but a real one. Of course, in practice it's not much likely that container does not have services and/or parameters, but technically speaking it's possible 🙂.

ondrejmirtes:
Thank you.

PR #422, opened 2025-01-09
feat: added stub for DataMapperInterface

JefvdA:
Ensure TData of FormInterface is provided
Set TData on mixed since we don't know the type of the child forms Add template for  in DataMapperInterface

creds to @roerbakei ;)

refs: #417

### DIFF
--- a/stubs/Symfony/Component/Form/DataMapperInterface.stub
+++ b/stubs/Symfony/Component/Form/DataMapperInterface.stub
@@ -1,0 +1,22 @@
+<?php
+
+namespace Symfony\Component\Form;
+
+/**
+ * @template TViewData
+ */
+interface DataMapperInterface
+{
+    /**
+     * @param TViewData|null $viewData
+     * @param \Traversable<mixed, FormInterface<mixed>> $forms
+     */
+    public function mapDataToForms(mixed $viewData, \Traversable $forms): void;
+
+    /**
+     * @param \Traversable<mixed, FormInterface<mixed>> $forms
+     * @param TViewData|null $viewData
+     * @param-out TViewData $viewData
+     */
+    public function mapFormsToData(\Traversable $forms, mixed &$viewData): void;
+}


Tomsgu:
This looks good to me. The only missing part is detecting if I pass a correct value when creating a form. Because currently I can pass anything I want.

```php
$form = $this->createForm(SnippetType::class, $notAllowedObject); // type mismatch error
```
```php
 /**
 * @template-implements DataMapperInterface<SnippetDto>
 */
class SnippetType extends AbstractType implements DataMapperInterface
{
    public function mapDataToForms(mixed $viewData, Traversable $forms): void
    {
        if ($viewData === null) {
            return;
        }

        // Does not make sense to do this check. But we still need to do it. Obvious error: 
        // Instanceof between SnippetDto and SnippetDto will always evaluate to true.
        // 🪪  instanceof.alwaysTrue
        
        //if (!$viewData instanceof SnippetDto) {
        //    throw new UnexpectedTypeException($viewData, SnippetDto::class);
        //}

        $forms = iterator_to_array($forms);
        // ...
    }
}
```
I am wondering if there is anything we can do about this.

ondrejmirtes (changes requested):
This is not sufficient. The stub has to be registered in extension.neon - stubFiles. Also because it's introducing generics, we need to add it to `skipCheckGenericClasses` like it was done here https://github.com/phpstan/phpstan-symfony/blob/dd1aaa7f85f9916222a2ce7e4d21072fe03958f4/extension.neon#L15-L25

ondrejmirtes:
There are two different PRs with this stub: https://github.com/phpstan/phpstan-symfony/pull/418

1) Both have the same mistake of not registering the stub.
2) Both are slightly different so you should agree first how it should look like.

stubs/Symfony/Component/Form/DataMapperInterface.stub
```diff
@@ -0,0 +1,22 @@
...
+    /**
+     * @param TViewData|null $viewData
+     * @param \Traversable<mixed, FormInterface<mixed>> $forms
+     */
+    public function mapDataToForms(mixed $viewData, \Traversable $forms): void;
+
+    /**
+     * @param \Traversable<mixed, FormInterface<mixed>> $forms
+     * @param TViewData|null $viewData
+     * @param-out TViewData $viewData
```
stof, line 19: This is wrong. there is no guarantee that the reference argument is replaced by a non-null value by the end of the method (and actually, the 3 core implementations of the interface have cases where the data will be `null` in the output)

PR #423, opened 2025-01-17
using messenger HandleTrait as QueryBus with appropriate result typing

bnowak:
This PR goal is to determine correct result typing for QueryBus kind of classes which uses SF messenger HandleTrait internally and simply return its results.

### DIFF
--- a/README.md
+++ b/README.md
@@ -232,20 +232,20 @@ class QueryBus
 
     public function dispatch(object $query): mixed
     {
-        return $this->handle($query); // Return type will be inferred
+        return $this->handle($query); // Return type will be inferred in calling code as query result
     }
 
     // Multiple methods per class example
     public function execute(object $message): mixed
     {
-        return $this->handle($message);
+        return $this->handle($message); // Return type will be inferred in calling code as query result
     }
 }
 
 // Interface-based configuration example
 interface QueryBusInterface
 {
-    public function dispatch(object $query): mixed;
+    public function dispatch(object $query): mixed; // Return type will be inferred in calling code as query result
 }
 
 class QueryBusWithInterface implements QueryBusInterface
@@ -263,7 +263,7 @@ class QueryBusWithInterface implements QueryBusInterface
     }
 }
 
-// Usage examples with proper type inference
+// Examples of use with proper type inference
 $query = new GetProductQuery($productId);
 $queryBus = new QueryBus($messageBus);
 $queryBusWithInterface = new QueryBusWithInterface($messageBus);
@@ -271,4 +271,5 @@ $queryBusWithInterface = new QueryBusWithInterface($messageBus);
 $product = $queryBus->dispatch($query); // Returns: Product
 $product2 = $queryBus->execute($query); // Returns: Product
 $product3 = $queryBusWithInterface->dispatch($query); // Returns: Product
+// Without the feature all above query bus results would be default 'mixed'.
 ```
--- a/README.md
+++ b/README.md
@@ -29,6 +29,7 @@ This extension provides following features:
 * Provides correct return type for `Extension::getConfiguration()` method.
 * Provides correct return type for `CacheInterface::get()` method based on the callback return type.
 * Provides correct return type for `BrowserKitAssertionsTrait::getClient()` method.
+* Provides configurable return type resolution for methods that internally use Messenger `HandleTrait`.
 * Notifies you when you try to get an unregistered service from the container.
 * Notifies you when you try to get a private service from the container.
 * Notifies you when you access undefined console command arguments or options.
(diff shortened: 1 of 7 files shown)


tests/Type/Symfony/data/messenger_handle_trait.php
```diff
@@ -111,3 +111,32 @@ public function __invoke()
...
     }
 }
+
+class QueryBus {
+    use HandleTrait;
+
+    public function dispatch(object $query): mixed
+    {
+        return $this->handle($query);
+    }
```
michaljusiega, line 67: We could also add this test case with changed function name.

```suggestion
    use HandleTrait {
        handle as private handleQuery;
    }

    public function handle(object $query): mixed
    {
        return $this->handleQuery($query);
    }
```

michaljusiega:
@ondrejmirtes Sorry to disturb you, but I think this is one of the pieces that is missing in this whole puzzle to support. Can we move forward ?

bnowak:
Hi @michaljusiega, thanks for your reply. I think we can add it - no problem. However, from my perspective it's more important to address the root issue (missing feature) of the topic (described [here](https://github.com/phpstan/phpstan-symfony/issues/424)). My reasoning is without having this result interpreted on bus class level - this is much less useful in general.

ondrejmirtes:
So this QueryBus class is something in your codebase and you want the same type inference that classes using HandleTrait and calling the `handle` method have?

You can simply write a dynamic return type extension for `QueryBus::dispatch()` that delegates the type resolution to `new MethodCall($methodCall->var, 'handle', $methodCall->getArgs())`.

bnowak:
@ondrejmirtes It's ready for review now. You were right about my intentions, however I wanted to make it working for any class that uses `HandleTrait` internally. I didn't want to hardcoded it, so made it parameterized. I've added also support for interfaces. Please review in you free time :pray: 

It's likely to be conflicted with my another [PR](https://github.com/phpstan/phpstan-symfony/pull/448) about splitting tests, so once first of these PR will be merged I can resolve conflicts on another one. :wink:

bnowak:
@ondrejmirtes any chances to review & merge it soon, please? :)

README.md
```diff
@@ -168,3 +170,94 @@ Call the new env in your `console-application.php`:
...
+```php
+use Symfony\Component\Messenger\Attribute\AsMessageHandler;
+
+// Product handler that returns Product
+#[AsMessageHandler]
+class GetProductQueryHandler
+{
+    public function __invoke(GetProductQuery $query): Product
+    {
+        return $this->productRepository->get($query->productId);
```
ondrejmirtes, line 212: Can you please add comments to the code in README about the inferred types with and without the extension? I'm still not sure what it does.

ondrejmirtes (changes requested):

README.md
bnowak: Thanks for reply.

So it's about inferring types of query bus classes (from CQRS pattern) or any custom class which internally use messenger's HandleTrait to get and return result of any dispatched message. Without the extension phpstan is not able to recognize that and gives generic return type of bus class (which in fact is some more specified query result). 

So the main difference is calling query bus classes, where based on passed/dispatched message we're getting its correct result type.

I've updated README a bit and the the direct outcome is shown [in the end of it](https://github.com/phpstan/phpstan-symfony/pull/423/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R255-R263) - it shows that we're getting `Product` type instead of generic `mixed`. If that's still not precised and unclear, please advice me what else I could add/adjust. Maybe my describing skills are not so good :sweat_smile:

bnowak: @ondrejmirtes any chances to re-look on this please?

bnowak: @ondrejmirtes ping :pray:

bnowak: @ondrejmirtes ping2 🙏

bnowak: @ondrejmirtes please, could you take a look on this again? and sorry for bothering you so many times

bnowak:
@ondrejmirtes could you review it again please?

src/Type/Symfony/MessengerHandleTraitWrapperReturnTypeExtension.php
```diff
@@ -0,0 +1,125 @@
...
+
+		$args = $expr->getArgs();
+		if (count($args) !== 1) {
+			return null;
+		}
+
+		$arg = $args[0]->value;
+		$argClassNames = $scope->getType($arg)->getObjectClassNames();
+
+		if (count($argClassNames) === 1) {
```
VincentLanglet: Instead of a check count === 1.

Can't you do
```
if (count($argClassNames) === 0) {
     return null;
}

$returnTypes = [];
foreach ($argClassNames as $argClassName) {
     $messageMap = $this->getMessageMap();
     $returnType = $messageMap->getTypeForClass($argClassNames[0]);

	if ($returnType === null) {
			return null;
     }
     $returnTypes[] = $returnType;
}

return TypeCombinator::union(...$returnTypes);
```
?

The goal would be to support union like
```
$queryBus = rand(0, 1) ? $productBus : $itemBus;
$productOrItem = $queryBus->dispatch($query); // Returns: Product|Item
```

src/Type/Symfony/MessengerHandleTraitWrapperReturnTypeExtension.php
```diff
@@ -0,0 +1,125 @@
...
+	{
+		if (!($expr instanceof MethodCall) || !($expr->name instanceof Identifier)) {
+			return false;
+		}
+
+		$methodName = $expr->name->name;
+		$varType = $scope->getType($expr->var);
+		$classNames = $varType->getObjectClassNames();
+
+		if (count($classNames) !== 1) {
```
VincentLanglet: Can't you do something like
```
if (count($classNames) === 0) {
     return false;
}

foreach ($classNames as $className) {
       if (!$this->isClassNameSupported($className)) {
            return false;
       }
}

return true;
```

In order to support
```
$query = rand(0, 1) ? $query1 : $query2;
$product = $queryBus->dispatch($query);
```

src/Type/Symfony/MessengerHandleTraitWrapperReturnTypeExtension.php
bnowak: :+1: adjusted

src/Type/Symfony/MessengerHandleTraitWrapperReturnTypeExtension.php
bnowak: :+1: adjusted

bnowak:
@VincentLanglet thank you for your reply and review. I've addressed your comments and also added support for union types in classes which use `HandleTrait` directly (not only on their wrappers/query buses).

Please take a look and re-review.

Will that change be automatically merged into 2.x branch as well when approved? :crossed_fingers:

VincentLanglet:
Oh, didn't see you were targeting the 1.4.x branch.

You should target the 2.0.x instead ; there is no feature added to the version 1.

tests/Type/Symfony/data/messenger_handle_trait.php
```diff
@@ -50,7 +50,69 @@ public function __invoke()
...
+
         // HandleTrait will throw exception in fact due to multiple handle methods/handlers per single query
         assertType('mixed', $this->handle(new MultiHandlersForTheSameMessageQuery()));
     }
 }
+
+class QueryBus {
+    use HandleTrait;
+
+    public function dispatch(object $query): mixed
```
VincentLanglet: To avoid the failure on php 7.4 I think you have to replace `: mixed` by a phpdoc `@return mixed` (everywhere in this file)

bnowak: thanks for advice. I've tried that, but with just adding phpdoc it didn't change anything. It started working when I simply removed native return type (it's mixed by default).

Now it should work (I've tested locally on PHP 7.4 & 8.4), but now one CI job failed (on  github.com authentication). I don't see button for retrying it (perhaps I don't have such permissions).

@VincentLanglet could you take a look and rerun CI jobs, please?

bnowak:
@VincentLanglet rebased to 2.0.x branch, but there's still issue with failed CI job (mentioned [here](https://github.com/phpstan/phpstan-symfony/pull/423#discussion_r2822914790)).

Could you take a look on that and re-review once again? I hope, I already covered everything in this PR :pray:

tests/Type/Symfony/data/messenger_handle_trait.php
VincentLanglet: I restarted the Ci it's ok

VincentLanglet (approved):

VincentLanglet:
Since you initially requested changes @ondrejmirtes, your review is required here :)

bnowak:
Thank you @VincentLanglet!

@ondrejmirtes kindly request to take a look at that🙏

ondrejmirtes (approved):

ondrejmirtes:
Thank you!

PR #435, opened 2025-03-19
Synchronize the EventSubscriberInterface with the upstream type

stof:
As of Symfony 5.4, Symfony ships a precise type for this interface but the stub overrides it.


ondrejmirtes:
Thank you.

PR #441, opened 2025-05-08
Avoid reporting with TestContainer::get as internal

VincentLanglet:
Hi @ondrejmirtes 

I tried to solve both https://github.com/phpstan/phpstan-symfony/pull/210 and https://github.com/symfony/symfony/issues/60299 with another implementation (by trying to copy the idea of the QueryType in phpstan-doctrine).

I wonder if it's something going in the right direction or not.


src/Rules/Symfony/ContainerInterfacePrivateServiceRule.php
```diff
@@ -43,11 +44,17 @@ public function processNode(Node $node, Scope $scope): array
 
 		$argType = $scope->getType($node->var);
 
-		$isTestContainerType = (new ObjectType('Symfony\Bundle\FrameworkBundle\Test\TestContainer'))->isSuperTypeOf($argType);
+		if (
+			$argType instanceof TestContainerType
```
ondrejmirtes: Could we simplify this and not report an error when we simply have the upper-most ContainerInterface and are in a subclass of KernelTestCase?

The custom type is wrong since it doesn't implement equals, describe and isSuperTypeOf. But I think we don't need it at all.

ondrejmirtes (changes requested):

src/Rules/Symfony/ContainerInterfacePrivateServiceRule.php
VincentLanglet: > Could we simplify this and not report an error when we simply have the upper-most ContainerInterface and are in a subclass of KernelTestCase?

I think this is not fully true, but this might be good enough.

VincentLanglet: Should be better with the last commit.

ondrejmirtes:
Thank you.

PR #443, opened 2025-05-20
Improve ParameterBag stub

### LINKED ISSUE #439: `ParameterBag->keys()` return type is incorrect
In the [`ParameterBag` stub](https://github.com/phpstan/phpstan-symfony/blob/2.0.x/stubs/Symfony/Component/HttpFoundation/ParameterBag.stub#L13) the return type is improperly narrowed to `list<string>`.

The implementation use a simple [`array_keys` call](https://github.com/symfony/http-foundation/blob/7.2/ParameterBag.php#L54). Therefore the return type MUST be `list<array-key>`.

With the current type narrowing it's impossible to correctly parse a query string with numeric keys (ie: `https://domain.tld/search?0=value&78=something`). Worse, trusting PHPStan than the key is a string might lead to a runtime `TypeError`.

nreynis:
Fixes: https://github.com/phpstan/phpstan-symfony/issues/439

### DIFF
--- a/stubs/Symfony/Component/HttpFoundation/ParameterBag.stub
+++ b/stubs/Symfony/Component/HttpFoundation/ParameterBag.stub
@@ -3,12 +3,12 @@
 namespace Symfony\Component\HttpFoundation;
 
 /**
- * @implements \IteratorAggregate<string, mixed>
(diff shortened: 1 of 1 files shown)


ondrejmirtes:
Thank you.

PR #447, opened 2025-07-10
Test Symfony 7.3

ondrejmirtes:


martin3398:
Is there any news on updating to newer versions of symfony. Right now we are getting errors like `XML /app/container.xml does not contain container.services.service structure` when we run this plugin. Even though we generated an XML container

ondrejmirtes:
@martin3398 Please open an issue and create a small reproducing repository if possible. I'm not aware about any incompatibilities with Symfony 7.3, your problem probably isn't exclusive to that.

PR #452, opened 2025-08-30
Add BrowserKitAssertionTraitReturnTypeExtension

VincentLanglet:


src/Type/Symfony/BrowserKitAssertionTraitReturnTypeExtension.php
```diff
@@ -0,0 +1,68 @@
...
+
+		return null;
+	}
+
+	/**
+	 * @phpstan-assert-if-true =MethodCall $expr
+	 */
+	private function isSupported(Expr $expr, Scope $scope): bool
+	{
+		if (!($expr instanceof MethodCall) || !($expr->name instanceof Identifier) || $expr->name->name !== self::TRAIT_METHOD_NAME) {
```
ondrejmirtes: I don't understand the logic. You're interested in the method name and you're interested where we're calling it from.

But you're not interested on what we're calling the method. This extension might be influencing a total coincidence that's nothing to do with `BrowserKitAssertionsTrait::getClient()`.

We're in a class, the class has `getClient` method. And the method is declared in BrowserKitAssertionsTrait trait. But the method we're still calling might be unrelated.

My hunch is telling me you're no checking `$scope->getMethodReflection($scope->getType($expr->var), $expr->name->name)` but you should be.

Also, method names are case-insensitive, so toLowerString() needs to happen.

ondrejmirtes (changes requested):

src/Type/Symfony/BrowserKitAssertionTraitReturnTypeExtension.php
VincentLanglet: I copied the logic from the extension you gave me as example
https://github.com/phpstan/phpstan-symfony/blob/8820c22d785c235f69bb48da3d41e688bc8a1796/src/Type/Symfony/MessengerHandleTraitReturnTypeExtension.php#L67-L87

So I assume it should be improve on the other extension too.

VincentLanglet:
I tried to fix both BrowserKitAssertionTraitReturnTypeExtension and MessengerHandleTraitReturnTypeExtension then @ondrejmirtes

VincentLanglet:
Friendly ping if you have time for a new look @ondrejmirtes

ondrejmirtes (approved):

ondrejmirtes:
Thank you!

PR #456, opened 2025-10-07
Fix "Possibly invalid array key type string|null."

staabm:
fixes build error

```
  ------ ---------------------------------------------- 
  Line   src/Symfony/XmlServiceMapFactory.php          
 ------ ---------------------------------------------- 
  86     Possibly invalid array key type string|null.  
         🪪  offsetAccess.invalidOffset                
 ------ ---------------------------------------------- 
 ```


ondrejmirtes:
Thank you!

PR #457, opened 2025-10-16
Allow psr/container ^2.0 in composer dependencies

muamadev:


ondrejmirtes:
Why?

muamadev:
> Why?

Hello,
I use `phpstan-symfony` and `GotenbergBundle`, they use the `psr/container `but with different version, I submitted this PR to avoid the conflict.

See. https://github.com/sensiolabs/GotenbergBundle/blob/v1.0.0/composer.json#L26 :)
Thanks

ondrejmirtes:
Require-dev section packages do not cause conflicts when someone is installing this package. You can have psr/container in any version and still use phpstan-symfony.

muamadev:
> Require-dev section packages do not cause conflicts when someone is installing this package. You can have psr/container in any version and still use phpstan-symfony.

You're right, my issue were caused by another package.
Thanks

PR #459, opened 2025-11-10
Support stringNode in Config component

emodric:
Symfony 7.2 added `>stringNode()` method to `NodeBuilder`.


src/Type/Symfony/Config/ArrayNodeDefinitionPrototypeDynamicReturnTypeExtension.php
```diff
@@ -19,6 +19,7 @@ final class ArrayNodeDefinitionPrototypeDynamicReturnTypeExtension implements Dy
 	private const PROTOTYPE_METHODS = [
 		'arrayPrototype',
 		'scalarPrototype',
+        'stringPrototype',
```
staabm: Indent seems off (also in other files)

staabm (commented):
Misses a test

emodric:
@staabm I've fixed the indentation issue and I've tried adding a test, but it turns out that tests do not support Symfony 7.

I'll leave this up once this lib gets up to Symfony 7.

stof:
In Symfony 7.4, we have conditional return types in Symfony itself that should make this return type extension useless. It might be better to disable it for new Symfony versions to benefit from the upstream work.

PR #460, opened 2025-11-23
Fix Serializer return type for arrays of primitives

yoeunes:
Fixes type inference for primitive scalar types (`int`, `string`, `bool`, `float`) when used with `Serializer::deserialize()`.

Currently, the extension wraps all base types in `ObjectType`. This means `int[]` is incorrectly inferred as an array of objects named "int" rather than scalar integers, causing false positives during arithmetic operations.


yoeunes:
While working on this fix, I noticed we could simplify and future-proof this code by using PHPStan's native `TypeStringResolver` instead of manual parsing.

Proposed refactoring:

```php
$types[] = $this->typeStringResolver->resolve($constantString->getValue());
```

I kept the minimal fix here, because changing the constructor to inject the resolver would be a **backward compatibility break**

PR #467, opened 2026-01-20
Fix Docblock for InputBag

### LINKED ISSUE #453: InputBag::get() method incorrectly reports type error for bool default parameter
PHPStan is incorrectly reporting a type error when using `false` as the default parameter for `Symfony\Component\HttpFoundation\InputBag::get()` method. 

The error states that the method expects `string|null` but `false` is given, which is incorrect according to the actual method signature.

PHPStan reports the following error:
```
Parameter #2 $default of method Symfony\Component\HttpFoundation\InputBag<string>::get() expects string|null, false given.
```

Code Example

```php
use Symfony\Component\HttpFoundation\Request;

$request = new Request();
$isPreview = $request->query->get('preview', false); // This triggers the error
```

Environment Details
- PHPStan Version: 1.12.25
- PHPStan Symfony Extension: 1.4.15
- Symfony Version: 6.4.*
- PHP Version: 8.2.20

The actual `InputBag::get()` method signature is:
```php
/**
 * Returns a scalar input value by name.
 *
 * @param string|int|float|bool|null $default The default value if the input key does not exist
 */
public function get(string $key, mixed $default = null): string|int|float|bool|null
{
    if (null !== $default && !\is_scalar($default) && !$default instanceof \Stringable) {
        throw new \InvalidArgumentException(sprintf('Expected a scalar value as a 2nd argument to "%s()", "%s" given.', __METHOD__, get_debug_type($default)));
    }

    $value = parent::get($key, $this);

    if (null !== $value && $this !== $value && !\is_scalar($value) && !$value instanceof \Stringable) {
        throw new BadRequestException(sprintf('Input value "%s" contains a non-scalar value.', $key));
    }

    return $this === $value ? $default : $value;
}
```

This appears to be an issue with PHPStan's Symfony extension having an outdated or incorrect type definition for the InputBag::get() method. The extension seems to be using an older type definition that doesn't match the current Symfony implementation.

HypeMC:
Fixes #453

The doc block was copied from the original class: https://github.com/symfony/symfony/pull/61698

### DIFF
--- a/stubs/Symfony/Component/HttpFoundation/InputBag.stub
+++ b/stubs/Symfony/Component/HttpFoundation/InputBag.stub
@@ -3,14 +3,15 @@
 namespace Symfony\Component\HttpFoundation;
 
 /**
- * @template TValue of string|int|float|bool
+ * @template TInput of string|int|float|bool|null
  */
 final class InputBag extends ParameterBag
 {
 
 	/**
-	 * @param TValue|null $default
-	 * @return TValue|null
+     * @template TDefault of string|int|float|bool|null
+     * @param TDefault $default
+     * @return TDefault|TInput
 	 */
 	public function get(string $key, $default = null)
 	{


ondrejmirtes:
Thank you!

PR #470, opened 2026-02-11
Fix phpstan error about generics missingType

VincentLanglet:


staabm (approved):
I wonder whether we could infer the generic types from the const array, like we do for other data-providers with phpstan-phpunit

PR #479, opened 2026-04-21
Remove useless stubs for Symfony console

stof:
Those stubs (added in #229) don't provide any improvements over the upstream types. Testing a return type extension does not require defining a stub for that method.


stof:
Closing this. Those stubs are relevant for Symfony 5.4 which is still supported by the extension, so it will be handled in #431 instead.

PR #480, opened 2026-04-21
Remove unnecessary ServiceSubscriberInterface stub

stof:
This stub does not override any upstream types and is not necessary for the stubbed methods of AbstractController.


stof:
As discussed in the Symfony Slack `#static-analysis` channel, stub classes don't need to implement the interfaces that are implemented in the actual interface (they are partial stubs after all), unless they reference them in `@implements` to provide the generic types (which is not the case here)

VincentLanglet:
Hi, can you rebase ; thanks

stof:
@VincentLanglet done

VincentLanglet (approved):

Issue #430, opened 2025-03-12
Form APIs should not be enforcing generic types that have been rejected in Symfony because of being wrong


stof:
Generic types for `FormTypeInterface` and the various `createForm` methods have been rejected in Symfony because they are lying about the actual behavior of the form component. Those generic types allow to have a nice inferred type for the basic usage of the component, but they will analyze advanced usages of the component in a broken way (reporting errors on valid code, inferring wrong types due to making wrong assumptions about how the component actually works).
See https://github.com/symfony/symfony/pull/40783#issuecomment-942470623 for the discussion when those generic types were submitted upstream.

In version 1.x of the extension, those generic types were configured in `skipCheckGenericClasses`, which allowed projects to mostly ignore the fact that the extension stubs were adding it. In 2.0x, this is not the case anymore so phpstan will enforce that any form type defines such generic type to satisfy the extension, even when being aware that those generic types are a lie.

My opinion is that those stubs should be removed to match the decision done in Symfony. However, I would also accept a case where the loading of those simplistic generic types is conditional, allowing projects to decide whether they want them (if they rely only on the basic usage pattern where the inference would produce a good result) or not (if they want the analysis to stick to the true behavior of the component).

ondrejmirtes:
These should be conditionally loaded in a StubFilesExtension based on the Symfony packages versions. We already do that in phpstan-doctrine with doctrine/collections I think.

stof:
This comment seems to be an answer for #431 rather than for this one.

For the form API, my point is that generic types added in those stubs are lying about the actual behavior of the component, which will mark some valid code as invalid and some invalid code as valid. The analysis I made in Symfony made me conclude that we cannot accurately represent the form behavior with generic types, especially not inside FormTypeInterface (which has to deal with a totally mutable form builder, and in some parts with the Form instance that is impacted by _all_ form types involved in building it)

ondrejmirtes:
Please show some code samples where the current form stubs are correct, and when they are wrong, so I can think about this more. My thinking is that they might still be useful for the "happy path", and some projects only use those happy paths, so removing the stubs would be a loss for some.

stof:
> My thinking is that they might still be useful for the "happy path", and some projects only use those happy paths, so removing the stubs would be a loss for some.

that's why I proposed to make their loading configurable (either by using a dedicated extension or by using a parameter in that extension).

> Please show some code samples where the current form stubs are correct, and when they are wrong, so I can think about this more.

[...]

Issue #400, opened 2024-06-27
Type refinement for options in form types


stof:
the `$options` argument of the various methods in FormTypeInterface and FormTypeExtensionInterface are typed as `array<string, mixed>` for now, because there is no other way to type them in phpdoc. However, the actual type is a configurable array shape, depending on the configuration of the OptionsResolver.
Having a more precise type would make it easier to write form types when using the level 9 of phpstan (where `mixed` requires explicit type checking to use them) and would allow more precise analysis at lower levels (as phpstan would report invalid usages of values taken from options instead of allowing anything because they are `mixed`)

Calls done on the OptionsResolver in the `configureOptions` method should be analyzed to build the array shape:
- `setDefined` marks a key (or multiple ones) as known (but the key might be missing)
- `setRequired` will mark a key (or multiple ones) as known and the key is guaranteed to be present
- `setDefault` will mark the key as known and the key is guaranteed to be present
- `setDefaults` is equivalent to calling `setDefault` for each key/value pair of the argument
- `setAllowedTypes` defines the allowed types for that option. When multiple values are passed in the array, this allows a union of those types. Supported types are built-in type names, object types (class names or interface names) or generic array types (defines with a `[]` suffix after a supported type). Passing a string as second argument is equivalent to passing an array containing that string as single value.
- `addAllowedTypes` allows to expand the list of allowed types by adding more types in the union (if no allowed types have been configured yet for that option, either through `setAllowedTypes` or `addAllowedTypes`, it is equivalent to `setAllowedTypes`)
- `setAllowedValues` allows to define the list of allowed values for a property (passing anything else than an array as second argument is treated like passing a single-value array):
    - either a `Closure` (not other types of callables) taking the value as argument and returning a boolean to decide whether it is allowed (which would probably mean that phpstan has to skip the refinement based on allowed values, unless it is able to refine the value for simple callables by reusing some logic supporting `assert()` type refinements)
    - or a value which will be compared with `===`. This might lead to refining the option to a union of string literals for instance.
- `setDefault` (and `setDefaults`) support defining a nested option by using a Closure with a first argument using `Symfony\Component\OptionsResolver\OptionsResolver` (exactly) as argument type. the type of the value will depend on the array shape for the OptionsResolver configured in that Closure:
    - if `setPrototype(true)` was called in the configuration, the nested option contains an array of values matching the array shape
    - otherwise, the nested option will match the array shape

[...]

PR #481, opened 2026-04-21
Add a dynamic stub loader to prefer upstream types

### LINKED ISSUE #431: Stubs should not be loaded when using versions of Symfony with upstream generic types

stof:
Once the precise types are added in Symfony, the upstream types should be used to benefit from the maintenance work of the Symfony team.

Closes #431


stof:
The remaining stubs belong to a few categories:

- the form stubs, which will be handled in #430 as we rejected them upstream
- stubs that have not yet been submitted upstream (mostly about some array iterable value types). I plan to submit them upstream and then follow up by making the loading conditional in phpstan-symfony based on the version including them
- the `Options` stub that might not make sense to move upstream as is, but might be useful to implement #400 when inferring the option resolver shape
- stubs necessary only as dependencies of other stubs

src/Stubs/Symfony/StubFilesExtensionLoader.php
```diff
@@ -0,0 +1,155 @@
...
+			$files[] = $stubsDir . '/Twig/Node/Node.stub';
+		}
+
+		return $files;
+	}
+
+	private function isInstalledVersionBelow(string $package, string $maxVersion): bool
+	{
+		if (!class_exists(InstalledVersions::class)) {
+			return false;
```
VincentLanglet, line 182: In which case this will occur ?

Should we return true, just to avoid a BC break ?

src/Stubs/Symfony/StubFilesExtensionLoader.php
```diff
@@ -0,0 +1,155 @@
...
+		}
+
+		if ($this->isInstalledVersionBelow('symfony/http-foundation', '6.2.0.0')) {
+			$files[] = $stubsDir . '/Symfony/Component/HttpFoundation/Cookie.stub';
+		}
+
+		if ($this->isInstalledVersionBelow('symfony/http-foundation', '7.4.0.0')) {
+			$files[] = $stubsDir . '/Symfony/Component/HttpFoundation/ParameterBag.stub';
+
+			try {
```
VincentLanglet, line 82: Can't we check that the version is >= some version ?

src/Stubs/Symfony/StubFilesExtensionLoader.php
stof: I think this could happen when using a phpstan phar without loading the project composer autoloader (which likely causes issues). Otherwise, it should not happen when using composer 2+.

I copied this condition from the stub loader of phpstan-doctrine.

src/Stubs/Symfony/StubFilesExtensionLoader.php
stof: I could. I copied the way it was done in the existing code that I moved here (to avoid having to handle coordinated conditions between 2 loaders).
Tell me whether you want me to do the change).

For reference, InputBag shipped in version 5.1.0 of the component.

stof:
Most remaining stub types (i.e. all except form types and the `Options` class) have been submitted upstream in https://github.com/symfony/symfony/pull/64009.
I'll prepare a follow-up PR to load those types conditionally, to be merged once the Symfony PR gets accepted.

src/Stubs/Symfony/StubFilesExtensionLoader.php
VincentLanglet: Didn't see it was coming from another file

VincentLanglet (approved):

extension.neon
```diff
@@ -11,29 +11,13 @@ parameters:
 		messenger:
 			handleTraitWrappers: []
 	stubFiles:
-		- stubs/Psr/Cache/CacheException.stub
-		- stubs/Psr/Cache/CacheItemInterface.stub
-		- stubs/Psr/Cache/InvalidArgumentException.stub
 		- stubs/Symfony/Bundle/FrameworkBundle/Controller/AbstractController.stub
```
staabm, line 14: what is the case for files which still are referenced in extension.neon?

are these required for every symfony version?

stof: See https://github.com/phpstan/phpstan-symfony/pull/481#issuecomment-4289911444 where I explain that

stof:
@VincentLanglet As the Symfony PR is merged, do you prefer that I update this PR with the conditional loading for Symfony 8.1.0 or keeping it as a follow-up PR (as this PR is already approved) ?

VincentLanglet:
> @VincentLanglet As the Symfony PR is merged, do you prefer that I update this PR with the conditional loading for Symfony 8.1.0 or keeping it as a follow-up PR (as this PR is already approved) ?

I'm fine with both. I wait for the approval from staabm anyway

stof:
@VincentLanglet @staabm I pushed my commit handling the stubs covered in Symfony 8.1 then (I already had it ready locally).

VincentLanglet (approved):

staabm (approved):

PR #483, opened 2026-04-22
Add support for newer Symfony versions in tests

stof:
Symfony 5.4 is in security-only mode since years (and would already be EOL if its maintenance window had not been extended compared to the normal release schedule). Testing only Symfony 5.4 in CI does not reflect current usage of the extension.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -26,17 +26,17 @@
     "phpstan/phpstan-phpunit": "^2.0.8",
     "phpstan/phpstan-strict-rules": "^2.0",
     "phpunit/phpunit": "^9.6",
-    "psr/container": "1.1.2",
-    "symfony/config": "^5.4 || ^6.1",
-    "symfony/console": "^5.4 || ^6.1",
-    "symfony/dependency-injection": "^5.4 || ^6.1",
-    "symfony/form": "^5.4 || ^6.1",
-    "symfony/framework-bundle": "^5.4 || ^6.1",
-    "symfony/http-foundation": "^5.4 || ^6.1",
-    "symfony/messenger": "^5.4",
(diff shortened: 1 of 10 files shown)


ondrejmirtes:
I started working on something similar here but it's challenging https://github.com/phpstan/phpstan-symfony/pull/447/changes

VincentLanglet:
This would be great ! Thanks for working on it

stof:
@VincentLanglet @ondrejmirtes what is the way to handle those `function.alreadyNarrowedType` errors reported by phpstan in places where tests have to alter their behavior based on the version of Symfony components being installed ? The reported errors will depend on the jobs (`lowest` jobs and jobs for older PHP versions will install some older versions of components, that will still have the old deprecated methods, or might not have the newer APIs, depending on the case).
I see that https://github.com/phpstan/phpstan-symfony/pull/447/changes#diff-a03dfdc240bbe0ba15bbac79aee5cb3182b00e9cfde3f5549d455acbd4f762d1 uses `new ReflectionClass()->hasMethod()` instead if `method_exists`, which does not seem to trigger narrowing in phpstan for now. Maybe that's an option for now (until phpstan implements such narrowing).

for the test failures themselves, I will handle them (tests relying on the messenger MessageSubscriberInterface must be skipped for versions where the interface does not exist anymore, and we should have alternative tests using the `#[AsMessageHandler]` attribute instead)

VincentLanglet:
What about a conditional baseline, like it's done on PHPStan repository with php versions
https://github.com/phpstan/phpstan-src/blob/2.2.x/build/ignore-by-php-version.neon.php
?

But I'm fine with a phpstan-ignore comment

PR #484, opened 2026-04-22
Fix the tests running the extension without the container

stof:
PHPunit expects test classes to have a `Test` suffix in their name by default. Other classes are not loaded (considering they are utilities or fixtures for the tests instead).

### DIFF
--- a/tests/Type/Symfony/ExtensionWithoutContainerTest.php
+++ b/tests/Type/Symfony/ExtensionWithoutContainerTest.php
@@ -5,7 +5,7 @@ namespace PHPStan\Type\Symfony;
 use PHPStan\Testing\TypeInferenceTestCase;
 use function class_exists;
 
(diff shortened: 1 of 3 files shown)


tests/Type/Symfony/ExtensionWithoutContainerTest.php
```diff
@@ -14,7 +14,7 @@ public function dataExampleController(): iterable
 			return;
 		}
 
-		yield from $this->gatherAssertTypes(__DIR__ . '/data/ExampleController.php');
+		yield from $this->gatherAssertTypes(__DIR__ . '/data/ExampleControllerWithoutContainer.php');
```
stof, line 17: Those were using the files defining assertions with precise container-based types instead of the dedicated assertions. This was not caught because the test never actually ran due to the class name not making it a test class.

VincentLanglet (approved):

VincentLanglet:
Thanks

PR #486, opened 2026-05-07
Remove Reflector from StubFilesExtensionLoader, replace with version checks

xificurk:
The loader previously used BetterReflection's Reflector to check whether InputBag existed before loading its stubs. This triggered SourceLocator initialization (an expensive cache-populating operation) on every analysis run.

Replace the reflectClass() call with an isInstalledVersionBelow() version check (InputBag was introduced in symfony/http-foundation 5.1). Add a test that wires a ThrowingSourceLocator into the DI container's betterReflectionSourceLocator service, so any future regression that causes getFiles() to touch BetterReflection will fail the test explicitly.

--

Related:
- https://github.com/phpstan/phpstan-src/pull/5577
- https://github.com/phpstan/phpstan-src/pull/5538#issuecomment-4342781825

### DIFF
--- a/src/Stubs/Symfony/StubFilesExtensionLoader.php
+++ b/src/Stubs/Symfony/StubFilesExtensionLoader.php
@@ -4,8 +4,6 @@ namespace PHPStan\Stubs\Symfony;
 
 use Composer\InstalledVersions;
 use OutOfBoundsException;
-use PHPStan\BetterReflection\Reflector\Exception\IdentifierNotFound;
-use PHPStan\BetterReflection\Reflector\Reflector;
 use PHPStan\PhpDoc\StubFilesExtension;
 use function class_exists;
 use function dirname;
@@ -13,15 +11,6 @@ use function version_compare;
 
 class StubFilesExtensionLoader implements StubFilesExtension
 {
-
-	private Reflector $reflector;
-
-	public function __construct(
-		Reflector $reflector
-	)
-	{
-		$this->reflector = $reflector;
-	}
 
 	public function getFiles(): array
 	{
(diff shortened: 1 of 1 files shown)


VincentLanglet:
For reference https://github.com/phpstan/phpstan-symfony/pull/481/changes#r3118949398

VincentLanglet (approved):

staabm:
I think it would be useful if we could have a phpbench here (similar as in phpstan-src), which makes sure changes in the extension will not regress performance in the future (actually such a benchmark would be useful for all the 1st party extensions).

how can the improvement of this change be reproduced?

xificurk:
@staabm I have updated the regression test to more closely cover the real use case, which is to avoid BetterReflection initialization just for ResultCache.

BetterReflection cache init cost becomes significant only on larger code base. I'm not sure how useful would phpbench be to guard this case - the dummy codebase would need to be big to avoid false positive regression signals.

tests/Analyser/ResultCacheNoReflectionTest.php
```diff
@@ -0,0 +1,36 @@
...
+
+namespace PHPStan\Analyser;
+
+use LogicException;
+use PHPStan\Analyser\ResultCache\ResultCacheManagerFactory;
+use PHPStan\BetterReflection\Reflector\Reflector;
+use PHPStan\Command\Output;
+use PHPStan\Testing\PHPStanTestCase;
+
+final class ResultCacheNoReflectionTest extends PHPStanTestCase
```
staabm: thanks for this test. 

@ondrejmirtes do you think this one would make sense directly in phpstan-src?
that way we could assert that ResultCache construction will not trigger reflection.

this assumption might be useful as general logic, and is not specific to phpstan-symfony

staabm: based on the test added by @xificurk here, I added one in phpstan-src with https://github.com/phpstan/phpstan-src/pull/5617

I think this is more useful, as we can later on cover all phpstan 1st party extensions with a single test

staabm:
this got approved by vincent, so I will merge this with the fix-only to verify the newly added phpstan-src test works as expected.

thank you @xificurk

ondrejmirtes:
Please don't forget to tag the release 😊

staabm:
ok cool. I was waiting for a signal that this is fine for you. thanks

PR #487, opened 2026-05-08
Improve laziness of container XML parsing

xificurk:
Parsing the XML into the DTOs is a costly operation on large Kernels. This PR defers the ServiceMap and ParameterMap initialization until the first use of their content.

### DIFF
--- a/src/Symfony/LazyParameterMap.php
+++ b/src/Symfony/LazyParameterMap.php
@@ -1,0 +1,58 @@
+<?php declare(strict_types = 1);
+
+namespace PHPStan\Symfony;
+
+use PhpParser\Node\Expr;
+use PHPStan\Analyser\Scope;
+
+abstract class LazyParameterMap implements ParameterMap
+{
+
+	private function __construct()
+	{
+	}
+
+	final public static function create(ParameterMapFactory $parameterMapFactory): self
+	{
+		// Workaround to make the static method getParameterKeysFromNode() work without sharing state across all LazyParameterMap instances
+		$lazyParameterMap = new class () extends LazyParameterMap
+		{
+
+			protected static ParameterMapFactory $parameterMapFactory;
+
+			protected static ?ParameterMap $parameterMap;
+
+			protected static function getParameterMap(): ParameterMap
+			{
+				self::$parameterMap ??= self::$parameterMapFactory->create();
+				return self::$parameterMap;
+			}
+
+		};
+		$lazyParameterMap::$parameterMapFactory = $parameterMapFactory;
+		$lazyParameterMap::$parameterMap = null;
+
+		return $lazyParameterMap;
+	}
+
+	abstract protected static function getParameterMap(): ParameterMap;
+
+	/**
+	 * @return ParameterDefinition[]
+	 */
+	public function getParameters(): array
+	{
+		return static::getParameterMap()->getParameters();
+	}
+
+	public function getParameter(string $key): ?ParameterDefinition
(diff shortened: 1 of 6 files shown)


src/Symfony/ServiceMap.php
```diff
@@ -18,6 +18,6 @@ public function getServices(): array;
 
 	public function getService(string $id): ?ServiceDefinition;
 
-	public static function getServiceIdFromNode(Expr $node, Scope $scope): ?string;
+	public function getServiceIdFromNode(Expr $node, Scope $scope): ?string;
```
VincentLanglet: I think this is a BC break, given the fact that the interface has `@api`. If someone implements the interface his code will break.

Same exist for ParameterMap ; so I'll need @ondrejmirtes review.

ondrejmirtes:
Yeah, removing the `static` keyword is a BC break. For 3.0 we could add `@api-do-not-implement` there (many PHPStan core interface already have it) and make the interface nicer. But I think this improvement could still be implemented without changing the interfaces?

xificurk:
@ondrejmirtes Yes, I'll rework this to maintain BC.

Removing the static would make sense for 3.0 - all the existing calls are on a dynamically resolved instance. No true static calls exist and given the implemented logic, they should not exist.

staabm:
@xificurk please also provide a profiler screenshot / hyperfine perf numbers so we can see the impact of this PR

src/Symfony/ServiceMap.php
xificurk: Rewritten to keep BC

xificurk:
@staabm As discussed in https://github.com/phpstan/phpstan-src/pull/5538 I am targetting few file partial analysis on larger code base.

A test on a very simple file `bin/phpstan analyse bin/console.php` median of 10 runs: 9.22s -> 8.77s. This is due to a complete elimination of ServiceMap and ParameterMap initialization in the worker process (confirmed via SPX trace).

staabm:
> median of 10 runs: 9.22s -> 8.77s.


Awesome 🤩

src/Symfony/LazyParameterMap.php
```diff
@@ -0,0 +1,58 @@
...
+
+			protected static function getParameterMap(): ParameterMap
+			{
+				self::$parameterMap ??= self::$parameterMapFactory->create();
+				return self::$parameterMap;
+			}
+
+		};
+		$lazyParameterMap::$parameterMapFactory = $parameterMapFactory;
+		$lazyParameterMap::$parameterMap = null;
```
staabm, line 33: can we move this into a `__construct` of the above inline class?

same in LazyServiceMap

xificurk: Not as long as LazyParameterMap::__construct remains private. The private constructor is intentional - it prevents any 3rd party inheritance.

staabm (approved):

ondrejmirtes:
> median of 10 runs: 9.22s -> 8.77s

What is the remaining 8 seconds taken by?

ondrejmirtes (approved):

xificurk:
> > median of 10 runs: 9.22s -> 8.77s
> 
> What is the remaining 8 seconds taken by?

Still mostly BetterReflection cache init, then there https://github.com/phpstan/phpstan-symfony/pull/488. Let me clear the obvious bottlenecks, and let's continue this in https://github.com/phpstan/phpstan-src/pull/5538

PR #488, opened 2026-05-09
SymfonyContainerResultCacheMetaExtension optimization

xificurk:
The cache busting strategy implemented by the extension attempts to avoid unnecessary invalidations by calculating the hash from the parsed parameters and services used within phpstan-symfony logic. This addresses the case when the container XML have been regenerated and changed, but not in an significant way impacting the result cache (originally discussed in https://github.com/phpstan/phpstan-symfony/issues/255).

The problem is that on large DI containers this parsing, normalizing and hashing is not exactly cheap:

Scrn_20260509_103902

This PR adds a simple filesytem cache for the calculated hash to avoid this cost - the hash is recalculated from the parsed services/parameters only when the underlying XML changes.

### DIFF
--- a/src/Symfony/SymfonyContainerResultCacheMetaExtension.php
+++ b/src/Symfony/SymfonyContainerResultCacheMetaExtension.php
@@ -4,6 +4,7 @@ namespace PHPStan\Symfony;
 
 use PHPStan\Analyser\ResultCache\ResultCacheMetaExtension;
 use function array_map;
+use function basename;
 use function file_get_contents;
 use function file_put_contents;
 use function hash;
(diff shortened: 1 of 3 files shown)


staabm:
@Wirone @ruudk you both contributed into the discussion of this feature. could you have a look at this proposal?

@xificurk how much does this change/improve your real world project? could you send some before/after 'hyperfine' numbers (measured without profiler-extension overhead)?

xificurk:
Measured on top of https://github.com/phpstan/phpstan-symfony/pull/487#issuecomment-4413097233

In combination with the lazy maps, this PR eliminates the initialization of ServiceMap and ParameterMap in the main process as well, leading to further improvement (10 run median)  8.77s -> 8.18s on the same simple test `bin/phpstan analyse bin/console.php`

src/Symfony/SymfonyContainerResultCacheMetaExtension.php
```diff
@@ -28,6 +38,29 @@ public function getKey(): string
...
+		if ($this->containerXmlPath !== null) {
+			$xmlHash = hash_file('sha256', $this->containerXmlPath);
+			if ($xmlHash === false) {
+				throw new XmlContainerNotExistsException(sprintf('Container %s does not exist', $this->containerXmlPath));
+			}
+			$hashFile = sprintf('%s/%s-%s.hash', $this->tmpDir, $this->getKey(), $xmlHash);
+			$hash = @file_get_contents($hashFile);
+			if ($hash !== false) {
+				return $hash;
+			}
```
ruudk: Do I understand correctly that this wil create new .hash files for every new hash of the XML contents? So that could result in 100's of .hash files?

It feels inefficient to first hash the file, to find the .hash file? Why not store this in a static .hash file that always contains the hash of the file?

xificurk: There are two hashes we need to keep track of:
1) the hash of the XML file content (for a cheap check whether the container has been re-generated)
2) the hash for result cache busting calculated from the service/parameter maps

Currently (1) is used in .hash filename, (2) is stored inside that file.

Good point with the possibility of the hundreds of stale .hash files - fixed by 741b799

ruudk (approved):

staabm:
atm we are focused on shipping the next phpstan release. I will comeback to this PR after the release is out

staabm (changes requested):
> The cache busting strategy implemented by the extension attempts to avoid unnecessary invalidations by calculating the hash from the parsed parameters and services used within phpstan-symfony logic. This addresses the case when the container XML have been regenerated and changed, but not in an significant way impacting the result cache 

I think we need a test in which the container is changed in a non-significant way, but the cache is still used

xificurk:
> > The cache busting strategy implemented by the extension attempts to avoid unnecessary invalidations by calculating the hash from the parsed parameters and services used within phpstan-symfony logic. This addresses the case when the container XML have been regenerated and changed, but not in an significant way impacting the result cache
> 
> I think we need a test in which the container is changed in a non-significant way, but the cache is still used

@staabm I can take a look at that, but want to be clear that what you quoted is the description of the current behavior present in the master. This MR is a performance optimization of that described strategy.

staabm (approved):
I see. I am fine than. thanks for noting.

VincentLanglet (approved):

staabm:
thanks anyone inolved!

PR #489, opened 2026-05-09
Update release-toot.yml

staabm:
copied this change from the phpstan repo.. maybe this fixes the toot error on releases

grafik

Todo
- [ ] port to other repos, after we verified it works


staabm:
since the previous action is already broken, lets merge and see whether it improves things on the next release :)

PR #497, opened 2026-06-04
Precise Uid::toString return type

VincentLanglet:
Related to https://github.com/symfony/symfony/pull/64450

This require to bump the lower version of PHPStan unless PHPStan has a new version which parse `non-decimal-int-string` and `decimal-int-string` as `string`.

### DIFF
--- a/tests/Type/Symfony/data/uid.php
+++ b/tests/Type/Symfony/data/uid.php
@@ -1,0 +1,14 @@
+<?php
+
+namespace UuidStub;
+
+use Symfony\Component\Uid\Uuid;
+use function PHPStan\Testing\assertType;
+
+$uuid = new Uuid::v4();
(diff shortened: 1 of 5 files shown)


staabm:
even if its just a stub change, I think it would be useful to see a nsrt test to get a real world example for this

tests/Type/Symfony/data/uid.php
```diff
@@ -0,0 +1,14 @@
+<?php
+
+namespace UuidStub;
+
+use Symfony\Component\Uid\Uuid;
+use function PHPStan\Testing\assertType;
+
+$uuid = new Uuid::v4();
```
staabm, line 8: typo?

```suggestion
$uuid = Uuid::v4();
```

PR #499, opened 2026-06-09
Default FormType data to mixed

alexander-schranz:
The same form type can be used with an `object` or if no data is set unstructured array `array<string, mixed>` or a none compound simple type `string`, `bool`, ...

So the best default to not annoy here is set TData to mixed, still all existing can say they only accept specific data.

In most cases form types not accessing data in them so this mostly unused generic type. Only a few form types accessing data directly.

### DIFF
--- a/stubs/Symfony/Component/Form/AbstractType.stub
+++ b/stubs/Symfony/Component/Form/AbstractType.stub
@@ -3,7 +3,7 @@
 namespace Symfony\Component\Form;
 
 /**
- * @template TData
+ * @template TData = mixed
  *
  * @implements FormTypeInterface<TData>
  */
--- a/stubs/Symfony/Component/Form/FormBuilderInterface.stub
+++ b/stubs/Symfony/Component/Form/FormBuilderInterface.stub
@@ -3,7 +3,7 @@
 namespace Symfony\Component\Form;
 
 /**
- * @template TData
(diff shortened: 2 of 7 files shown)


VincentLanglet (approved):

staabm (commented):
I have not much experience with these types.

feel free to merge if you guys know what you are doing :)

VincentLanglet:
> I have not much experience with these types.
> 
> feel free to merge if you guys know what you are doing :)

I think the goal here is to  add a default in order to avoid forcing the user to type every existing generic Form ; especially because those generic are partially wrong (see https://github.com/phpstan/phpstan-symfony/issues/430)

alexander-schranz:
Yes `mixed` is here better as making wrong assumption during the extension forcing a generic type here. Didn't see #430 before but I agree here that the Form Component is to complex and the phpstan extension forces wrong assumption for generic types now. 

So better set a default to mixed and not force them.  
Still a form type which set a specific `data_class` and `empty_data` and validates it to be only that kind of type can override the generic type still built of top of them.

But default to mixed make it easier, specially you not require accessing the data in lot of custom types. I even would consider it be bad practice outside of listener but thats only my personal opinion :). As mention there also be can data transformers which form component allow via Extension manipulate every type.

PR #503, opened 2026-07-17
Do not resolve ServiceLocator get()/has() against the global container

janedbal:
`ServiceDynamicReturnTypeExtension` and `ContainerInterfaceUnknownServiceRule` are registered for `Psr\Container\ContainerInterface`, which also matches an injected `Symfony\Component\DependencyInjection\ServiceLocator` (e.g. from `#[AutowireLocator]`). A locator's keys are **local index keys** (tag index attributes), not global container service ids — but both consulted the global `ServiceMap`.

Consequences for a `ServiceLocator`:

- `has('key')` collapsed to a constant `false` (the key isn't a global service id), so `if (!$locator->has('key')) { return; }` became *"always true"* and the rest of the method **unreachable**;
- `get('key')` was wrongly reported with `symfonyContainer.serviceNotFound`.

This PR skips `ServiceLocator` receivers in both places, leaving `get()`/`has()` to native (generic) inference. The global container, controllers, and Psr-typed `ServiceSubscriber` locators are unaffected (a `ServiceLocator` is not their static type; the existing `testGetPrivateServiceInLegacyServiceSubscriber` still passes).

Discovered via shipmonk-rnd/dead-code-detector#376 — the unreachable code led a dead-code detector to report the handler methods behind the locator as unused. Repro & analysis: https://github.com/shipmonk-rnd/dead-code-detector/issues/376#issuecomment-5002396100

### DIFF
--- a/src/Rules/Symfony/ContainerInterfaceUnknownServiceRule.php
+++ b/src/Rules/Symfony/ContainerInterfaceUnknownServiceRule.php
@@ -50,6 +50,14 @@ final class ContainerInterfaceUnknownServiceRule implements Rule
 			return [];
 		}
 
+		// A ServiceLocator is a scoped locator: its keys are local (e.g. tag index
+		// attributes from #[AutowireLocator]), not global container service ids, so the
+		// ServiceMap has no knowledge of them and cannot tell whether they are registered.
+		$isServiceLocatorType = (new ObjectType('Symfony\Component\DependencyInjection\ServiceLocator'))->isSuperTypeOf($argType);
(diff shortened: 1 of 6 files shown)


pkly:
Technically speaking it's possible to resolve the actual values that exist by looking at the container xml.

The service might actually have the argument registered properly in the xml, which will resolve to a ServiceLocator id which can then be found in the container, which in turn could be used to look up actual values.

But... this would require reading the wired service declaration from the container, which could be impossible if the class is registered under an alias (in which case, skipping seems like the only "valid" option).

